PR #47, opened 2015-02-08
fixed parameters of scmVersion.tag documentation in README.md

tschulte:
When I tried to change the serialize and deserialize rules I found the documentation in README.md was wrong.


adamdubiel:
You are right, thanks for the contribution! FYI i am currently working on better documentation which will be hosted on readthedocs.

PR #49, opened 2015-02-09
1.2.0 - documentation and hooks

adamdubiel:


docs/Makefile
```diff
@@ -0,0 +1,177 @@
+# Makefile for Sphinx documentation
```
szpak: Wow, Makefile in Java project :).
Maybe there will  a Gradle plugin one day...

martinda:
I have found an answer in the new proposed documentation that I could not find in the current doc (about Jenkins). And I really like the more extensive doc (Version Creation, the "Reading" section in particular). This reinforces the need for me to check that releases are only made from a single branch in a single repository (in my case master). I don't want users to do ad hoc releases from random branches just because they can type "gradle release". Anyway, I am not qualified to do a review of this new code, just giving you some comments about what I like.

docs/Makefile
adamdubiel: I doubt, they would need gradle at readthedocs end :) This documentation is not build during axion-release build.

adamdubiel:
@martinda that's exactly why i moved to readthedocs, readme could not contain all information i wanted to share anymore. Glad you like it :)
As to the "gradle release from random branches" - we use release only on CI, so it is easy to control it.

martinda:
In the old documentation, there was a short section describing the main tasks of the plugin. In this description, the release task specifies that the tag is created and pushed to the remote. 
I could not find this information in the new doc.

martinda:
The old documentation also had a long example, which I find useful when I need to see all the features at a glance. Can you please bring that back in the new doc? Thanks in advance.

adamdubiel:
Okay, i will update the documentation before merge - thanks for the feedback and let me know if there is anything missing still.

adamdubiel:
@martinda - i added missing bits of documentation

martinda:
Thank you!

PR #51, opened 2015-02-15
Add Maven Central and JCenter badges

szpak:
To click on and get code snippet to put into Gradle/Maven build file.

Btw, there is no version in Bintray's badge. It looks ok, so I hope it will appear soon (it is probably better to wait for before merging that PR).
Btw2, I assume sync to Maven Central is still planned.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@ axion-release-plugin
 
 *gradle release and version management plugin*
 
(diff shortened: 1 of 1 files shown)


adamdubiel:
Yes, it has been pushed and now i'm waiting for Sonatype -> Central sync. Thanks!

Issue #70, opened 2015-05-01
axion should always release if force version


vbuell:
It's impossible to release forced version when:
* we are currently on released version but we want to release yet another version (say, we have many releasing branches and I want to release from the same position both 1.0.0 and 0.99.15)
* in case if something went wrong last time and axion already had created tag but later didn't managed to push changes to remote; and we want to re-release it again after we have fixed the issue

For cases like this it make sense to treat forceVersion option more strictly way - it should try to re-release the version if possible, even if we are already on released version.

vbuell:
Pull Request: https://github.com/allegro/axion-release-plugin/pull/73

adamdubiel:
Second case won't work - since we already have a tag with same name, git will complain (i guess it's #69 that fixes that). As to the first option - why would you like to have the same piece of code marked by two different versions? axion-release will return the latter tag always anyway. Can you draft an example, when this will actually work?

vbuell:
The realistic examples that happened in my practice:
- Releasing final version after the release candidate, if no bugs was fixed. Like: 2.0.0-rc5, 2.0.0-GA. Both may point to the same git commit.
- Sometimes you have to release something at the end of the spring (for sprint demoing, for smoke testing, etc). In that case 2.0.0-s47 - where s47 is the # on sprint - may point to the same git commit as 2.0.0.

Regarding the second case - you're absolutely right - #69 should care about that.

adamdubiel:
Okay, i see the point - please take a look at my code review comments and i will merge it afterwards.

adamdubiel:
Released in 1.2.4.

PR #73, opened 2015-05-01
Always release if force version

vbuell:
PR for https://github.com/allegro/axion-release-plugin/issues/70

== copy from Issue ==
It's impossible to release forced version when:

- we are currently on released version but we want to release yet another version (say, we have many releasing branches and I want to release from the same position both 1.0.0 and 0.99.15)
- in case if something went wrong last time and axion already had created tag but later didn't managed to push changes to remote; and we want to re-release it again after we have fixed the issue

For cases like this it make sense to treat forceVersion option more strictly way - it should try to re-release the version if possible, even if we are already on released version.

### DIFF
--- a/src/main/groovy/pl/allegro/tech/build/axion/release/domain/Releaser.groovy
+++ b/src/main/groovy/pl/allegro/tech/build/axion/release/domain/Releaser.groovy
@@ -1,6 +1,7 @@
 package pl.allegro.tech.build.axion.release.domain
 
 import com.github.zafarkhaja.semver.Version
+import org.gradle.api.Project
 import org.gradle.api.logging.Logger
 import pl.allegro.tech.build.axion.release.domain.hooks.ReleaseHooksRunner
 import pl.allegro.tech.build.axion.release.domain.scm.ScmService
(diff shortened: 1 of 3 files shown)


src/main/groovy/pl/allegro/tech/build/axion/release/domain/VersionConfig.groovy
```diff
@@ -96,11 +96,15 @@ class VersionConfig {
     VersionWithPosition getRawVersion() {
         if (rawVersion == null) {
             ensureVersionServiceExists()
-            rawVersion = versionService.currentVersion(this, VersionReadOptions.fromProject(project))
+            rawVersion = versionService.currentVersion(this, getVersionReadOptions())
         }
         return rawVersion
     }
 
+    VersionReadOptions getVersionReadOptions() {
```
adamdubiel: if you opt to create a helper method, keep it private (VersionConfig is available as `scmVersion` extension in Gradle script)

adamdubiel: ah, i see it in Releaser now, nevermind

adamdubiel: Okay, after giving it a bit of thought - if we carry on with this PR (see my comment for the issue), the `VersionConfig$getVersionReadOptions` should be removed/kept private and `Releaser` should remain unchanged. I don't want `VersionConfig` to expose anything more than absolute minimum (even if it means crossing the DRY).

vbuell:
your comment is addressed and rebased with master

PR #75, opened 2015-05-03
Code clean-up.

vbuell:
+ Code clean-up.
+ Trim forceVersion value.
+ Added logging into .tag()


src/main/groovy/pl/allegro/tech/build/axion/release/infrastructure/git/GitRepository.groovy
```diff
@@ -18,6 +19,7 @@ import pl.allegro.tech.build.axion.release.domain.scm.*
 
 import java.util.regex.Pattern
 
+@Slf4j
```
adamdubiel: i would rather stick to Gradle project.logger thingy here

adamdubiel:
There is just one comment, other than that - it's perfect. Thanks for the cleanup :) i will merge as soon as you change it (please leave comment, cos otherwise i wont be notified of change).

vbuell:
Logger issue addressed

Issue #52, opened 2015-02-19
Working with git flow and develop branch


coco1979ka:
I am using the git flow model to manage my repository. Unfortunately this leads to a problem with the currentVersion task. All my work is going to the development branch. At some time I make a release, let's say 0.1.0. My master now contains a Tag 0.1.0. Now I start working for 0.2.0 in development.
First Problem:
Current Version in development branch still is 0.1.0-SNAPSHOT (which I can solve by setting a Tag with release-0.2.0-alpha)
Second Problem: 
I make a hotfix for version 0.1.0 which leads to a Tag release-0.1.1 in Master, but with the merge of the hotfix branch back to development branch, currentVersion now is 0.1.1-SNAPSHOT, because now the Tag of my first release also is known in the development branch. 
Is there a solution with a special versionCreator? 
![bildschirmfoto 2015-02-19 um 10 57 38](https://cloud.githubusercontent.com/assets/5807679/6264676/2765a3d8-b826-11e4-8838-b689fbb57731.png)

adamdubiel:
I see the problem. axion-gradle tries to find first tag in commit history, without evaluating if there were any  higher-version markers before this one. This behavior could be changed to:

* read current position
* am i on tag? okay, not changing anything
* not on tag and previous tag was release? search for closes nextVersion marker
* found nextVersion marker with Version higher than previous tag? use nextVersion instead

This would solve the problem and i think it's quite easy to implement.

coco1979ka:
I agree, but it only solves the second problem and implies manually setting the nextVersion marker. If it would be possible to search for a merge commit in history (usually this is automated commit message like "merge branch release/0.1.0"), it could be improved by:
* if on branch develop and none of the above conditions matches: If has merge release commit, increment minor version of merged version.
With this convention based version increment, it would only be necessary to set the next version marker, when a new minor release is planned.
So I would suggest a git flow version creator, which can be configured with a few options like:
* Development Branch Name
* Release merge search pattern
I could start to implement it and send you a pull request.

adamdubiel:
Do you mean something like i proposed in #50? :) Maybe we should merge those issues and come up with one solution.

coco1979ka:
Yes this looks quite similar. The feature should be implemented based on:
* http://jeffkreeftmeijer.com/2010/why-arent-you-using-git-flow/
* https://github.com/nvie/gitflow

Second link is a high level extension to git, which is very useful, as it handles merges automatically based on the current workflow step.

[...]

Issue #50, opened 2015-02-10
Abitlity to create a release branch and to use createReleaseCommit on it


martinda:
Some workflows involve the creation of a release branch. Would this be an appropriate feature for this plugin? A branch would be created, then a release commit would be made to that branch. IT would work along with the createReleaseCommit feature, in the sense that the commit would be made to the release branch. The release branch would be named in a similar manner as the tag is named. What do you think?

adamdubiel:
I think it's a nice feature, but we have to think about merging stuff. If release branch won't be merged to master/production, the version won't propagate there - do you have any thoughts on this?

martinda:
I was thinking of the workflow presented here http://blogs.atlassian.com/2013/10/inside-atlassian-feature-branching-on-the-stash-team/ (scroll down to the branching model picture).

Assuming the release number is structured as Major.minor.bugfix, release numbers would have to be based on the release branches, and on their tags. Release branches would need to be initialized (tagged) to start with a new Major release number (1.0.0, 2.0.0, ...). Feature releases would increment the minor number (1.1.0, 1.2.0, ...). Merges from bugfixes would increment the bugfix number (1.1.1, 1.1.2, ...). The plugin would need to have a few new tasks to increment specific elements of the release number depending on what is being done (feature, bugfix or new release stream). I am not sure whether this changes the search that gets the currentVersion value. It may not if the plugin searches from the current commit: the user or the CI system could prepare the local repo with the right checkout. I feel this problem is workflow specific and should not be the concern of the plugin.

For users of this workflow, it is harder to get a snapshot release number: a user would need to get (merge) their feature into a release branch before she could get a snapshot release number (when working locally, she could merge into master then into the feature branch, or directly into the feature branch).

I think the plugin needs to be workflow agnostic, but that it should not preclude any specific workflow. Workflow complications could be handled in the CI system and by the users who choose a more complicated workflow.

I think the plugin only need three more methods: one to increment each element of the version number. And it may have these methods already, I have not looked at the plugin code yet.

adamdubiel:
I will crunch this idea overnight, but quick comment: there is no way to increment custom part of version, but adding this is a really quick fix.

martinda:
I just found a gradle plugin that implements [Semantic Versionning](https://github.com/ajoberstar/gradle-git/wiki/SemVer%20Support) after following a grgit link.

adamdubiel:
[...]

PR #78, opened 2015-05-14
conditionalRelease task

AlfredBroda:
I created a conditionalRelease task for automatic release detection during continuous integration builds - specifically on Bamboo where string comparison is impossible in scripts. The assumption is that everytime you commit to the "releaseBranch" you should have a release version created. This task can be safely executed on any branch, it will only actually release on the specified branch. Configuration is as simple as placing the following code in the build.gradle file:
``` groovy
conditionalRelease {
    releaseBranch = "production"
}
```

### DIFF
--- a/src/main/groovy/pl/allegro/tech/build/axion/release/ConditionalReleaseTask.groovy
+++ b/src/main/groovy/pl/allegro/tech/build/axion/release/ConditionalReleaseTask.groovy
@@ -1,0 +1,27 @@
+package pl.allegro.tech.build.axion.release
+
+import org.gradle.api.DefaultTask
+import org.gradle.api.tasks.TaskAction
+import pl.allegro.tech.build.axion.release.domain.Releaser
+import pl.allegro.tech.build.axion.release.domain.scm.ScmRepository
+import pl.allegro.tech.build.axion.release.infrastructure.di.Context
+
+import java.util.regex.Pattern
+
+class ConditionalReleaseTask  extends DefaultTask {
+    String releaseBranch = "master"
+
+    @TaskAction
+    void conditionalRelease() {
+        println("Release branch is: " + releaseBranch)
+        Context context = new Context(project)
+        ScmRepository repository = context.repository()
+        def branch = repository.currentPosition(Pattern.compile(".*")).branch
+        if (branch == releaseBranch) {
+            Releaser releaser = context.releaser()
+
+            releaser.release(context.config())
+            releaser.pushRelease(context.config())
+        }
+    }
+}
--- a/src/main/groovy/pl/allegro/tech/build/axion/release/ConditionalReleaseTask.groovy
+++ b/src/main/groovy/pl/allegro/tech/build/axion/release/ConditionalReleaseTask.groovy
@@ -1,23 +1,20 @@
 package pl.allegro.tech.build.axion.release
-
 import org.gradle.api.DefaultTask
 import org.gradle.api.tasks.TaskAction
 import pl.allegro.tech.build.axion.release.domain.Releaser
 import pl.allegro.tech.build.axion.release.domain.scm.ScmRepository
 import pl.allegro.tech.build.axion.release.infrastructure.di.Context
 
-import java.util.regex.Pattern
-
 class ConditionalReleaseTask  extends DefaultTask {
     String releaseBranch = "master"
 
     @TaskAction
     void conditionalRelease() {
-        println("Release branch is: " + releaseBranch)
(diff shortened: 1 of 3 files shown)


src/main/groovy/pl/allegro/tech/build/axion/release/ConditionalReleaseTask.groovy
```diff
@@ -0,0 +1,27 @@
...
+import pl.allegro.tech.build.axion.release.infrastructure.di.Context
+
+import java.util.regex.Pattern
+
+class ConditionalReleaseTask  extends DefaultTask {
+    String releaseBranch = "master"
+
+    @TaskAction
+    void conditionalRelease() {
+        println("Release branch is: " + releaseBranch)
```
adamdubiel: hm, leftovers from debug? if not, use gradle logging

adamdubiel:
can you also add changes to rtd docs (./docs/..) ?

AlfredBroda:
Sure, I missed the docs somehow, will fix that. The printline is intentional, should it be logged at a different level? Info or debug for example?

src/main/groovy/pl/allegro/tech/build/axion/release/ConditionalReleaseTask.groovy
adamdubiel: log.info i suppose, but lifecycle might be okay as well

vbuell:
Usually you have few release branches and they looks like 'release/1.11'. Regular expression would be better.

`releaseBranch = "release/.*"`

src/main/groovy/pl/allegro/tech/build/axion/release/ConditionalReleaseTask.groovy
```diff
@@ -0,0 +1,27 @@
...
+        Context context = new Context(project)
+        ScmRepository repository = context.repository()
+        def branch = repository.currentPosition(Pattern.compile(".*")).branch
```
adamdubiel: you coudl use bit groovier syntax like `~/.../`

src/main/groovy/pl/allegro/tech/build/axion/release/ConditionalReleaseTask.groovy
```diff
@@ -0,0 +1,27 @@
...
+
+import org.gradle.api.DefaultTask
+import org.gradle.api.tasks.TaskAction
+import pl.allegro.tech.build.axion.release.domain.Releaser
+import pl.allegro.tech.build.axion.release.domain.scm.ScmRepository
+import pl.allegro.tech.build.axion.release.infrastructure.di.Context
+
+import java.util.regex.Pattern
+
+class ConditionalReleaseTask  extends DefaultTask {
```
adamdubiel: i miss some kind of test here - tasks are quite hard to test, so i move all logic into separate workers, thus eliminating all *logic* from tasks (thus releaser). maybe you could do the same?

adamdubiel:
+1 to @vbuell

vbuell:
Since the configuration of branch name pattern you introduced may be useful for other logic (see #52 or #79), would putting `releaseBranch` into root instead of `conditionalRelease` be more reusable?

AlfredBroda:
I'm still working on the documentation, I was thinking about putting my initial notes about configuration into the basic usage file, what do you think?

adamdubiel:
I think that basic usage should be left for really basic cases - the ones that get you started :) I would add some docs in Tasks as well as Overview. Then separate page in Configuration section.

vbuell:
I just created a PR that introduces the similar configuration value - https://github.com/allegro/axion-release-plugin/pull/82.
Basically, my concern is that the configuration of branch name pattern have to be shared for every features that must know the name of release branch - your story, #79, #52, #50. So, this configuration value should sit in top level rather than in `conditionalRelease`.

AlfredBroda:
True, I'll try to merge those and provide some documentation this weekend.

vbuell:
I'd say that we would have a little bit cleaner solution if not to stick to some specific case. What if to add a new configuration, say `activationTriggers`, that activates releasing using either predefined rules or closures:

```
conditionalRelease {
   activationTriggers: ['whenOnRelease', '<someOtherRule>']
}
```

Probably we need a negative set of rules also. For the scenario "Release when on master but not for 0.x.x versions" it would look like:

```
conditionalRelease {
   activationTriggers: ['whenOnRelease']
   suppressTriggers: [ { position -> position.startsWith("release/0.") } ]
}
```

adamdubiel:
closing due to inactivity

Issue #66, opened 2015-04-17
Release not working in Jenkins - Push fails with message "Nothing to push"


joergus:
Executing `gradle release` via Jenkins is not working for me. I'm using gradle 2.3 and the jenkins git plugin 2.3.5. 

I can reproduce that on a development machine when executing a checkout of a certain revision as it is done by jenkins `git checkout -f eced05a5302fa1bc452f57d3accc504965157344`.
I'm using the default git plugin.
The Exception is the following.
```
:release
Executing task ':release' (up-to-date check took 0.001 secs) due to:
  Task has not declared any outputs.
Working on released version 0.3.2, nothing to release.
Pushing all to remote: origin
The following authentication options are allowed (though they may not be available): [HARDCODED, SSHAGENT, PAGEANT, INTERACTIVE]
:release FAILED
:release (Thread[main,5,main]) completed. Took 0.478 secs.

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':release'.
> Problem pushing to remote.

* Try:
Run with --debug option to get more log output.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':release'.
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeActions(ExecuteActionsTaskExecuter.java:69)
	at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:46)
	at org.gradle.api.internal.tasks.execution.PostExecutionAnalysisTaskExecuter.execute(PostExecutionAnalysisTaskExecuter.java:35)
	at org.gradle.api.internal.tasks.execution.SkipUpToDateTaskExecuter.execute(SkipUpToDateTaskExecuter.java:64)
	at org.gradle.api.internal.tasks.execution.ValidatingTaskExecuter.execute(ValidatingTaskExecuter.java:58)
	at org.gradle.api.internal.tasks.execution.SkipEmptySourceFilesTaskExecuter.execute(SkipEmptySourceFilesTaskExecuter.java:42)
	at org.gradle.api.internal.tasks.execution.SkipTaskWithNoActionsExecuter.execute(SkipTaskWithNoActionsExecuter.java:52)
	at org.gradle.api.internal.tasks.execution.SkipOnlyIfTaskExecuter.execute(SkipOnlyIfTaskExecuter.java:53)
[...]

PR #81, opened 2015-05-18
Added repository configuration to only push tags to remote repositories

erichsend:
This relates to https://github.com/allegro/axion-release-plugin/issues/66

In a detached head state that can occur on some CI systems, the release task will fail when it attempts to push local commits to the remote repository. My attempt to fix was to add an optional configuration parameter to signal that only local tags, and not local commits, should be pushed to the remote repository.

Output of my jenkins job showing a successful push from a detached head state:
https://gist.github.com/erichsend/4b9327c58dec98e22ae2


src/main/groovy/pl/allegro/tech/build/axion/release/infrastructure/git/GitRepository.groovy
```diff
@@ -83,12 +87,16 @@ class GitRepository implements ScmRepository {
     }
 
     private void callPush(String remoteName, boolean all) {
-        repository.push(remote: remoteName, all: all)
+        if(pushTagsOnly == false) {
```
szpak: `!pushTagsOnly`?

src/main/groovy/pl/allegro/tech/build/axion/release/infrastructure/git/GitRepository.groovy
```diff
@@ -83,12 +87,16 @@ class GitRepository implements ScmRepository {
...
+            repository.push(remote: remoteName, all: all)
+        }
         repository.push(remote: remoteName, tags: true, all: all)
     }
 
     private void callLowLevelPush(ScmIdentity identity, String remoteName, boolean all) {
-        pushCommand(identity, remoteName, all).call()
+        if(pushTagsOnly == false) {
```
szpak: `!pushTagsOnly`?

szpak:
It would be probably good to have a simple test case for the new feature.

src/main/groovy/pl/allegro/tech/build/axion/release/infrastructure/git/GitRepository.groovy
erichsend: It's a style I picked up working with a programmer with bad eyes.

szpak: I agree that some people could prefer that style, but several lines earlier there is `if (!isOnExistingTag) {` and it would be probably good to keep similar style in the whole file. Of course there is a chance that `== false` is already used in some other places and maybe @adamdubiel will prefer to change everything to a verbose syntax.

erichsend: Indeed. I can only find one instance of '== false' in the codebase, and it's in a test. I'm changing it to the compact syntax to conform.

erichsend:
@szpak I agree. I am working on a test...just trying to figure out how and where to do it.

adamdubiel:
Waiting for the test before merge :) Could you also update the documentation? At least the overview section (in `docs/`).

erichsend:
@adamdubiel Sure thing. I just wanted to be sure this approach was acceptable before changing docs.

vbuell:
Adam, as Erich mentioned instead of introducing a new option we can just skip pushing if no 'releaseCommit' option specified.

src/main/groovy/pl/allegro/tech/build/axion/release/domain/scm/ScmInitializationOptions.groovy
```diff
@@ -12,23 +12,27 @@ class ScmInitializationOptions {
...
     final String remoteUrl
 
+    final boolean pushTagsOnly
+
     private static final String ATTACH_REMOTE_PROPERTY = 'release.attachRemote'
 
     private static final String FETCH_TAGS_PROPERTY = 'release.fetchTags'
 
-    ScmInitializationOptions(String remote, boolean fetchTags, boolean attachRemote, String remoteUrl) {
+    ScmInitializationOptions(String remote, boolean fetchTags, boolean attachRemote, String remoteUrl, boolean pushTagsOnly) {
```
adamdubiel: i wonder - this option might be set from cmd line as well (via `-Prelease.pusthTagsOnly`), so there would be no need to leak CI settings into buildscript

erichsend: I agree. I wasn't sure exactly where to pipe it in, I just needed to demonstrate the functionality. I was hoping you would have the better vision for how to apply it.

adamdubiel:
@vbuell - right, but then i think having manual override via cmdline for CI could be useful. The `pushTagsOnly` option needs to be added to Repository anyways, it's just question of how to set it's value. And then, it is not so easy to see if someone performs any commit in pre-release action, since she might use predefined action of `commit` type and not set `releaseCommit` option at all.

adamdubiel:
@erichsend wrapping it all up:

* i should be able to set `pushTagsOnly` from command line
* command line setting should take precedence over buildscript setting
* in general i think this flag should be evaluated as late as possible (meaning passed `repository#push`), allowing people to override it in runtime, but this would call for some refactoring and i consider it outside the scope of this pr

src/test/groovy/pl/allegro/tech/build/axion/release/infrastructure/git/GitRepositoryTest.groovy
```diff
@@ -220,6 +219,22 @@ class GitRepositoryTest extends Specification {
         remoteRawRepository.tag.list()*.fullName == ['refs/tags/release-push']
     }
 
+    def "should not push commits if the pushTagsOnly flag is set to true"() {
+        given:
+        initializationOptions = ScmInitializationOptions.fromProject(project, 'origin', true)
+        repository = new GitRepository(project.file('./repo'), identity, initializationOptions)
```
adamdubiel: hm, this creates nested repository, i would be careful with that - maybe just create new Project object? (this is a bit of hack for creating temp file, but it also makes the domain look nicer :P)

src/test/groovy/pl/allegro/tech/build/axion/release/infrastructure/git/GitRepositoryTest.groovy
```diff
@@ -220,6 +219,22 @@ class GitRepositoryTest extends Specification {
...
+
+        repository.tag('release-push')
+        repository.commit(['*'], 'commit after release-push')
+
+        when:
+        repository.push(ScmIdentity.defaultIdentity(), 'origin')
+
+        then:
+        remoteRawRepository.log(maxCommits: 1)*.fullMessage == ['InitialCommit']
```
adamdubiel: i'm bit confused and i don't know if this test tests the right thing - the new repository, created few lines up in `./repo` direcotry shouldn't have any remotes defined, only the "main" repository created in `setup` has this remote. Even if it somehow works as it should, for clarity i would like to have all definitions as explicit as possible, creating this repo from scratch would make the situation more understandable. Moving the flag into the `repository#push` would make this a lot easier :)

erichsend:
@adamdubiel 
Regarding the command line setting, should it be similar to dryRun, that is, if it is present then it is true? Or should it be able to be set as true or false from the command line property?

I currently have it set to behave like the dryRun flag:

```
     boolean pushTagsOnly = (project.hasProperty(REPOSITORY_PUSH_TAGS_ONLY_FLAG) ?
                true : config.pushTagsOnly)
```

erichsend:
@adamdubiel 
I believe this is ready. Let me know if you would like me to change the command line option to accept a true/false value instead of acting as an ON flag, or any other suggestions you have.

adamdubiel:
@erichsend - sorry for delay, i been on short vacation, i will take a look at this today or tomorrow

src/main/groovy/pl/allegro/tech/build/axion/release/infrastructure/di/ScmRepositoryFactory.groovy
```diff
@@ -14,14 +14,19 @@ class ScmRepositoryFactory {
 
     private static final String GIT = 'git'
 
+    private static final String REPOSITORY_PUSH_TAGS_ONLY_FLAG = "repository.pushTagsOnly";
```
adamdubiel: could you change the name to `release.pushTagsOnly`? Name clash is very unlikely for such a long flag, but i would like to keep everything in `release` namespace

adamdubiel:
There is just one "big" remark about naming. Other than that, i feel that there is room for improvement, but it's not critical and would involve bigger refactor. I will take care of it after merge. Please change the variable name and we are set.

erichsend:
@adamdubiel I agree this change is more of a work around than a good long term solution. Your suggestion of removing the flag as a property of the repository and instead adding it as an argument to the push() method certainly seems desirable. If I can find the time in the next week or two I could attempt that assuming you do not get to it first.

Another thing I would bring up is whether pushTagsOnly should be the default behavior. Maybe it is because my team has a policy to only do releases from the CI system, but I was actually surprised that the plugin was attempting to push local commits, as it seems like any commit important enough to be a release should first be published and vetted (by the CI system and peer review) before the release tag is applied. 

As I said though, my experience would be quite different from someone who primarily did all of their release maintenance from their laptop.

erichsend:
Regarding the naming, should the configuration item (not the flag) remain as a property in RepositoryConfig?

adamdubiel:
Naming: yest, this should be part of `repository` configuration, but on global level (gradle flags) it should be in `release.*` namespace.

As to the default behavior: we also release only from CI, but mind, that when using pre-release hooks there might be some local commits that i.e. change version number in Readme or some docs, that should be pushed as well. I would rather keep the new option optional instead of default not to break it :)

adamdubiel:
Merged, thanks for this contribution :)

erichsend:
Thanks for pulling this in, Adam, and thanks for the great plugin.

Issue #79, opened 2015-05-14
Custom incrementing rules


vbuell:
Axion should support custom semantic version incrementing strategies like:

- Increment patch version if on release branch. Increment minor version otherwise
- Increment pre-release version (rc2 -> rc3) if current version has pre-release part and it ends with digit. Increment patch version otherwise
- etc

adamdubiel:
I have a feeling solving this https://github.com/allegro/axion-release-plugin/issues/52 would cover also this usecase. In general the solution would be adding configurable version incrementer, that could decide how to change the version based on X (where X could be git history or current version).

btw sorry i haven't checked the DI code yet :(

vbuell:
Oh. I didn't see that. You're right, these two tightly coupled...  Will publish a PR soon.
Regarding DI - np. It's not a functional change, you know )

vbuell:
PR was accepted. Closing...

adamdubiel:
It took me long enough, but `1.3.0` is finally here! I changed the naming of configuration option to `versionIncrementer` and added additional rule `branchSpecific`, which delegates execution to other rule depending on branch. Everything is described in [docs](http://axion-release-plugin.readthedocs.org/en/latest/configuration/version.html#incrementing). Once again thanks for coding this feature :)

Issue #17, opened 2014-11-19
Treat project as new snapshot version when repo contains changed files


adamdubiel:
As a user, i want axion-release to treat repository with changed (but uncommited) files as next SNAPSHOT version. This should not take effect, when checks are disabled.

vbuell:
I'd say we should have even three levels of stability:

1. Treat as snapshot if not on tag (current behavior)
2. Treat as snapshot if not on tag or on tag with changed files (this issue)
3. Treat as snapshot if not on tag or on tag but the tag was created before this run. In other words, treat stable only for the build that is releasing now

The third case is mainly for CI jobs to avoid situations when new builds run of already released master branch are being treated as stable and trying to republish its artifacts. In general, the definition of stability of the project may be defined as an imaginary point on the timeline when the release was made (successfully); every consequent builds may or not be stable due to any side effects (dependencies, health of subsystems, etc) even if the codebase is intact.

adamdubiel:
I think the 3rd behavior should be managed using flag - we add/modify some files that are not part of repository during our CI build and i wouldn't want them to corrupt my version.

vbuell:
Agreed

vbuell:
Oh. I just realized that 3rd one probably is hardy doable due to the fact that a plugin does not allow to release and publish at the same run...

adamdubiel:
I found some time to fix this issue, it's in develop now. By default uncommitted changes are ignored and have no effect (backwards compatibility), but its easy to change:

```
scmVersion {
    ignoreUncommittedChanges = false
}
```

It can be always enabled by using flag:

```
./gradlew cV -Prelease.ignoreUncommittedChanges
```

Of course it's all in the docs as well.

PR #82, opened 2015-05-20
Support for custom version incrementing rules

vbuell:
For #79

### DIFF
--- a/docs/configuration/version.rst
+++ b/docs/configuration/version.rst
@@ -119,6 +119,35 @@ version::
 
 Input objects have same structure as deserialization closure inputs.
 
+.. _version-incrementing:
+
+Incrementing
+------------
+
+Incrementing phase does incrementing the version in accordance with *version incrementer* rule. By default version patch
+(least significant) number is incremented. There are predefined rules:
+
+- *incrementPatch* - increment patch number
+- *incrementMinor* - increment minor (middle) number
+- *incrementMinorIfNotOnRelease* - increment patch number if on release branch. Increment minor otherwise
+- *incrementPrerelease* - increment pre-release suffix if possible (-rc1 to -rc2). Increment patch otherwise
+
+You can set one of predefined rules via ``scmVersion.versionIncrementingRule`` method::
+
+    scmVersion {
+        versionIncrementingRule 'incrementPatch'
+    }
+
+Alternatively you can specify a custom rule by setting a closure that would accept a version, position and configuration
+and return a version object::
+
+    scmVersion {
+        versionIncrementingRule {version, position, versionConfig -> ...}
+    }
+
+If incrementMinorIfNotOnRelease rule is activated then the property ``releaseBranchPattern`` is used to match the release
+branch. By default it's set to 'release/.+'.
+
 .. _version-decorating:
 
 Decorating
--- a/docs/configuration/version.rst
+++ b/docs/configuration/version.rst
@@ -138,11 +138,11 @@ You can set one of predefined rules via ``scmVersion.versionIncrementingRule`` m
         versionIncrementingRule 'incrementPatch'
     }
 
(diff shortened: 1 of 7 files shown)


adamdubiel:
Thanks, i will take a look at it today or tomorrow :)

adamdubiel:
The code looks nice, i will run through it in Idea to see how things connect.
@szpak - we talked about something similar at Geecon, can you take a look at this pull request and see if it would solve your problem?

docs/configuration/version.rst
```diff
@@ -119,6 +119,35 @@ version::
...
+.. _version-incrementing:
+
+Incrementing
+------------
+
+Incrementing phase does incrementing the version in accordance with *version incrementer* rule. By default version patch
+(least significant) number is incremented. There are predefined rules:
+
+- *incrementPatch* - increment patch number
+- *incrementMinor* - increment minor (middle) number
```
szpak: There is an inconsistency between `incrementMinor` in a documentation and `incrementMinorVersion` in code. The same for `incrementPatch`.

vbuell: Oops. Good catch.
Fixed.

szpak:
@adamdubiel It's not a problem. It's a feature :).

It took a while to make it work, as my project was on a release tag and due to #17 I had no effect using the new parameter :). In the end I was able to do simple custom transformation as expected and after tweaking my code to do a Git lookup to parse the latest commit message in the `versionIncrementingRule` closure I should be able to set a new version using Git commit message.

Btw, guys, have you managed how to install plugin to a local repo with the new `plugin-publish-plugin`?

szpak:
Btw2, really nice work @vbuell !

vbuell:
Adam, all comments are addressed. Please review.

adamdubiel:
Sorry, i suffer from constant lack of time.. there is just one thing i would like to change. Imagine these dynamic version closures might require different arguments (like `incrementMinorIfNotOnRelease` needs `releaseBranchPattern`). I would not want to add a lot of these optional arguments. So maybe they should be configured:

* using map next to definition (e.x. `versionIncrementingRule 'incrementPatch' [ releaseBranchPattern: 'whatever*' ]`
* using parameters in closure, i.e. `versionIncrementingArguments {  releaseBranchPattern: 'whatever*' } ` , arguments in closure should be dynamic (like in map)

Personally i think the first option would be the best, this would also stay consistent with the way pre/post hooks are configured. Other than that - looks great :)

vbuell:
Done

adamdubiel:
Hm, i can still see this Pattern in ScmVersion and no Map for arguments of incrementer - can you tell me how do you understand my last comment? I like the introduction of Context object, it will be helpful in future.

vbuell:
Oh. I did not read it carefully. Sorry. Will redo.
In your example you suggest to move `releaseBranchPattern` from top-level. But if (when) you pull #78, release branch name will be duplicated in two places. I'd avoid that

adamdubiel:
Yeah, i was wondering about it as well.. can you make the change to be able to pass arguments *in general* and we can discuss what to do with this `releaseBranchPattern`.

vbuell:
It's not yagni :D

adamdubiel:
Oh, i already have use case for this :)

vbuell:
You intrigued me. :) Done

adamdubiel:
Thanks - i hope the build passes as well :) I'm going to spend some time with axion during the weekend, merging stuff, adding some small bugfixes and i would like to have 1.3.0 released on Sunday.

vbuell:
Forgot to say that I did not document this configuration map since there in no any predefined incrementers that use it yet. Let's wait till the first consumer of this map appeared

adamdubiel:
Exactly, that's fine.

vbuell:
Thanks for merging!

PR #106, opened 2015-10-18
Added flag to force SNAPSHOT version

### LINKED ISSUE #84: Custom logic for making a version "SNAPSHOT"
It'd be good, to have an option to override the default logic for determining whether the version should be marked as a snaphsot or not.
Preferably, it should work the similar way as overriding the versionCreator.

vbuell:
Resolves #84 and #17 (3rd behavior)


szpak:
LGTM. I like the use case with producing snapshots from a CI server.

adamdubiel:
Looks nice, thanks! Merging, i will give it a test run and release asap :)

PR #108, opened 2015-11-06
Upgrade to Gradle 2.8

martinda:
Adapt code to groovy 2.4

Gradle 2.8 upgrades groovy to 2.4
Groovy 2.4 has different operator precedence behavior


adamdubiel:
Thanks! :)

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Sat Jul 11 01:04:47 CEST 2015
+#Thu Nov 05 21:23:10 EST 2015
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.5-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.8-bin.zip
```
szpak: You could add also `distributionSha256Sum` which was added in Gradle 2.6  - https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:verification

martinda:
I went looking for the checksums, but Gradle does not have them: https://discuss.gradle.org/t/where-can-i-find-distribution-checksums/9404/1

szpak:
IT was implemented recently in Gradle 2.6 - https://docs.gradle.org/2.6/release-notes#support-for-verifying-gradle-wrapper-distribution-download-against-sha-256-hash

martinda:
Since the Gradle people do not publish the checksums, I had to do something a bit convoluted: I set the checksum to some wrong value, removed my local copy of ```gradle-2.8-bin.zip```, then run ```./gradlew tasks```, which downloads the ```gradle-2.8-bin.zip``` file. Since the expected checksum is wrong, Gradle issues an error message which contains the expected checksum and the actual checksum. I then changed the expected checksum in ```gradle-wrapper.properties``` to match the actual checksum, and it worked. Submitted as pull-request #110

szpak:
Hmm, couldn't you just calculate it manually as proposed in the first link I provided?

Anyway glad to see you managed to do it ;)

martinda:
I could calculate it manually, but then how do I know that my copy is valid? I am not the authoritative source. That's why it is important for Gradle to publish the checksums. It was luck that I found the convoluted method I described above!

szpak:
It would failed on Travis build when downloading a version from Gradle server.

But yes, I would be good to have official sha256 sums published and signed with the official Gradleware's pgp key.

PR #118, opened 2015-11-28
Provide a versionIncrementer called incrementMajor

martinda:
I have this usage model:
```
scmVersion {
  if (project.hasProperty('release.versionIncrementer')) {
    versionIncrementer project."release.versionIncrementer"
  }
}
```
I have added a PredefinedVersionIncrementer for the Major version, such that this command would work: ```./gradlew release -Prelease.versionIncrementer=incrementMajor```.

I do not know why a Major version incrementer is not part of the original plugin, I am new to this problem space.

### DIFF
--- a/docs/configuration/version.rst
+++ b/docs/configuration/version.rst
@@ -131,6 +131,7 @@ Incrementing phase does incrementing the version in accordance with *version inc
 
 * *incrementPatch* - increment patch number
 * *incrementMinor* - increment minor (middle) number
+* *incrementMajor* - increment major number
 * *incrementMinorIfNotOnRelease* - increment patch number if on release branch. Increment minor otherwise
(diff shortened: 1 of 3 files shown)


adamdubiel:
There is no spectacular reason, simply there was no demand for it :) I want yet to check the `-Prelease.versionIncrementer` as i don't think it is possible to use it at the moment. Could you add this new predefined incrementer where all others are documented? http://axion-release-plugin.readthedocs.org/en/latest/configuration/version.html#incrementing

adamdubiel:
Sorry for leaving it here for so long, i was working on major refactor and totally forgot about this PR, which could have been released a long time ago. I will merge it and also add a possibility to declare versionIncrementer via command line:

```
./graldew release -Prelease.versionIncrementer=incrementMajor
```

martinda:
Thank you!

PR #119, opened 2015-12-02
remove project.logger dependencies

zygm0nt:
I'd like to use code from this plugin in SBT build, but with Gradle Project references all over the place I just can't do this.

As a solution I've removed dependencies on `project.logger` in favour of Slf4j's Logger - as described here: https://docs.gradle.org/current/userguide/logging.html#sec:sending_your_own_log_messages

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,6 +18,7 @@ plugins {
 }
 
 apply plugin: 'com.gradle.plugin-publish'
+apply plugin: 'maven-publish'
 
 if(project.hasProperty('mavenCentral')) {
     apply from: 'gradle/mavenCentral.gradle'
@@ -101,5 +102,22 @@ pluginBundle {
     mavenCoordinates {
         groupId = 'pl.allegro.tech.build'
         artifactId = 'axion-release-plugin'
+    }
+}
+
+publishing {
+    publications {
+        mavenJava(MavenPublication) {
+            from components.java
+        }
+    }
+}
+
+publishing {
+    repositories {
+        maven {
+            // change to point to your repo, e.g. http://my.org/repo
+            url "/Users/marcin.cylke/.m2/repo"
+        }
     }
 }
--- a/src/main/groovy/pl/allegro/tech/build/axion/release/VerifyReleaseTask.groovy
+++ b/src/main/groovy/pl/allegro/tech/build/axion/release/VerifyReleaseTask.groovy
@@ -2,6 +2,8 @@ package pl.allegro.tech.build.axion.release
 
 import org.gradle.api.DefaultTask
 import org.gradle.api.tasks.TaskAction
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
 import pl.allegro.tech.build.axion.release.domain.ChecksResolver
 import pl.allegro.tech.build.axion.release.domain.LocalOnlyResolver
 import pl.allegro.tech.build.axion.release.domain.VersionConfig
@@ -10,6 +12,8 @@ import pl.allegro.tech.build.axion.release.domain.scm.ScmRepository
 import pl.allegro.tech.build.axion.release.infrastructure.di.Context
 
 class VerifyReleaseTask extends DefaultTask {
+
+    private final Logger logger = LoggerFactory.getLogger(VerifyReleaseTask)
 
     @TaskAction
     void prepare() {
@@ -25,7 +29,7 @@ class VerifyReleaseTask extends DefaultTask {
 
         if (resolver.checkUncommittedChanges()) {
             boolean uncommittedChanges = repository.checkUncommittedChanges()
-            project.logger.quiet("Looking for uncommitted changes.. ${uncommittedChanges ? 'FAILED' : ''}")
+            logger.trace("Looking for uncommitted changes.. ${uncommittedChanges ? 'FAILED' : ''}")
             if (uncommittedChanges && !dryRun) {
                 changesPrinter.printChanges()
 
@@ -33,18 +37,18 @@ class VerifyReleaseTask extends DefaultTask {
                         "See above for list of all changes.")
             }
         } else {
-            project.logger.quiet('Skipping uncommitted changes check')
+            logger.trace('Skipping uncommitted changes check')
         }
 
         boolean remoteAttached = repository.remoteAttached(config.repository.remote)
         if (resolver.checkAheadOfRemote() && !localOnlyResolver.localOnly(remoteAttached)) {
             boolean aheadOfRemote = repository.checkAheadOfRemote()
-            project.logger.quiet("Checking if branch is ahead of remote.. ${aheadOfRemote ? 'FAILED' : ''}")
+            logger.trace("Checking if branch is ahead of remote.. ${aheadOfRemote ? 'FAILED' : ''}")
             if (aheadOfRemote && !dryRun) {
                 throw new IllegalStateException("Current branch is ahead of remote counterpart - can't release.")
             }
         } else {
-            project.logger.quiet("Skipping ahead of remote check")
+            logger.trace("Skipping ahead of remote check")
         }
     }
 }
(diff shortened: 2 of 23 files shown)


src/main/groovy/pl/allegro/tech/build/axion/release/domain/scm/ScmInitializationOptions.groovy
```diff
@@ -23,12 +23,25 @@ class ScmInitializationOptions {
...
+        return [ATTACH_REMOTE_PROPERTY : extractOptionalParameter(project, ATTACH_REMOTE_PROPERTY),
+                FETCH_TAGS_PROPERTY : extractOptionalParameter(project, FETCH_TAGS_PROPERTY)]
+    }
+
+    private static def extractOptionalParameter(Project project, String paramName) {
+        return (String) (project.hasProperty(paramName) ? project.property(paramName) : null)
+    }
+
     static ScmInitializationOptions fromProject(Project project, String remote) {
+        return fromProject(extractParameters(project), remote)
```
vbuell: You can just use `fromProject(project.getProperties(), remote)`. There is no need to construct an intermediate map for holding these two options.

src/main/groovy/pl/allegro/tech/build/axion/release/domain/scm/ScmInitializationOptions.groovy
```diff
@@ -12,9 +12,9 @@ class ScmInitializationOptions {
 
     final String remoteUrl
 
-    private static final String ATTACH_REMOTE_PROPERTY = 'release.attachRemote'
+    static final String ATTACH_REMOTE_PROPERTY = 'release.attachRemote'
```
vbuell: Why package visibility?

src/main/groovy/pl/allegro/tech/build/axion/release/domain/scm/ScmInitializationOptions.groovy
```diff
@@ -23,12 +23,25 @@ class ScmInitializationOptions {
...
+
+    private static def extractOptionalParameter(Project project, String paramName) {
+        return (String) (project.hasProperty(paramName) ? project.property(paramName) : null)
+    }
+
     static ScmInitializationOptions fromProject(Project project, String remote) {
+        return fromProject(extractParameters(project), remote)
+    }
+
+    static ScmInitializationOptions fromProject(Map<String, ?> properties, String remote) {
```
vbuell: The name of this method doesn't represent what it actually does.

vbuell:
This PR introduces inconsistency - there are other places left where gradle's logging is used: VerifyReleaseTask, DryRepository, GradleAwareScmService, etc.

src/main/groovy/pl/allegro/tech/build/axion/release/domain/scm/ScmInitializationOptions.groovy
szpak: It is rather `public` after change, but in addition to the fact if it is needed somewhere else with the new code `ATTACH_REMOTE_PROPERTY` is not treated as a constant anymore in Groovy. A getter will be generated.

src/main/groovy/pl/allegro/tech/build/axion/release/infrastructure/git/GitRepository.groovy
```diff
@@ -14,14 +14,15 @@ import org.eclipse.jgit.revwalk.RevCommit
...
+import org.slf4j.LoggerFactory
 import pl.allegro.tech.build.axion.release.domain.scm.*
+import org.slf4j.Logger
 
 import java.util.regex.Pattern
 
 class GitRepository implements ScmRepository {
 
-    private final Logger log;
+    private final Logger log = LoggerFactory.getLogger(GitRepository.getClass())
```
szpak: Why not just `GitRepository.class`? Or as it is Groovy just `LoggerFactory.getLogger(GitRepository)`?

szpak:
@zygm0nt Do you plan to create an SBT plugin or just use the code in your project?

adamdubiel:
From what i understand they want to use this plugin from SBT. I asked the author for some more context, as i don't know how can Gradle plugin be used with SBT and what changes are needed. As you mentioned this PR introduces some inconsistency.

zygm0nt:
We have scala project with SBT build files in it. We would like to incorporate axion's version detection in it - to detect version from tags in a consistent manner with the rest of axion plugins' ecosystem. That is why I'd like to reuse the code from axion-release-plugin as much as I can. Ideally I'd like to just have hooks to specific functionalities of the plugin.

I haven't planned on creating sbt plugin, but it is actually a good idea. We are currently using sbt-release plugin and it is rather problematic. I'll create axion-like sbt plugin if the initial integration succeeds.

As for other remarks - I'll fix them on friday. 

If changing `project.logger` to slf4j logger is ok for you, than I can introduce this in other places.

adamdubiel:
@zygm0nt from what i understand in ideal situation we would have total separation between Gradle and version logic? This is actually not so hard to achieve, since i tried to keep Gradle-specific features fenced. Probably logging is the biggest leak here. I have started some work on separating runtime configuration from plugin configuration, which would make it even easier.

zygm0nt:
That's right! Although if you're aiming at moving gradle-independant code into separate project - this is not that important to me, and would also be troublesome during axion-release development.

adamdubiel:
If just a clean cut is enough then i would stop at it, without creating modules/projects :)

src/main/groovy/pl/allegro/tech/build/axion/release/domain/scm/ScmInitializationOptions.groovy
zygm0nt: No, I can't. I don't want gradle's Project leaking all over the place :)

src/main/groovy/pl/allegro/tech/build/axion/release/domain/scm/ScmInitializationOptions.groovy
zygm0nt: I changed scope to private.

src/main/groovy/pl/allegro/tech/build/axion/release/infrastructure/git/GitRepository.groovy
zygm0nt: Fixed.

src/main/groovy/pl/allegro/tech/build/axion/release/domain/scm/ScmInitializationOptions.groovy
vbuell: Huh?
But that class already deals with Project - it's one of argument of that method. It's already leaking )

build.gradle
```diff
@@ -18,6 +18,7 @@ plugins {
 }
 
 apply plugin: 'com.gradle.plugin-publish'
+apply plugin: 'maven-publish'
```
vbuell: Looks like it's a junk and shouldn't be committed

src/main/groovy/pl/allegro/tech/build/axion/release/VerifyReleaseTask.groovy
```diff
@@ -25,7 +29,7 @@ class VerifyReleaseTask extends DefaultTask {
 
         if (resolver.checkUncommittedChanges()) {
             boolean uncommittedChanges = repository.checkUncommittedChanges()
-            project.logger.quiet("Looking for uncommitted changes.. ${uncommittedChanges ? 'FAILED' : ''}")
+            logger.trace("Looking for uncommitted changes.. ${uncommittedChanges ? 'FAILED' : ''}")
```
vbuell: QUIET log level in gradle (despite its name) is higher than WARNING - it's always visible. You cannot map it to trace level which, I bet, nobody will see ever.

vbuell: That's BTW a big question here - there is no way to map quiet level to one of slf4j's levels. You cannot use warn or error because these messages are neither former nor latter. At the same time you cannot use info since there is no guarantee that it'll not be filtered out by logging configuration.

adamdubiel:
This should be fixed in #121

Issue #114, opened 2015-11-19
Different versioning schemes in maintenance branches


szpak:
I have a library which has maintenance/legacy branch for critical bugfixes for our key users and the master branch with current development. With branch specific `versionIncrementer` I can bump only a patch number, but it would be useful - as a complement to version increments and decorators - to have also a branch specific tag prefix (to not interfere between 2.x in master and 1.3.x in legacy). 

There is quite simple, yet not very elegant, workaround. In the legacy branch tag.prefix could be set to another value (e.g. legacy/) with current legacy version tag manually created (legacy/1.2.2). Therefore if you think it would complicate Axion configuration overly it is possible to live without it :).

adamdubiel:
Okay, i will take a look at this, though should not be very hard.

adamdubiel:
Well, its harder than expected, but i'm halfway done :) Though a lots of places have been touched and lots of tests to fix. On the plus side, no API changes except for adding new config option:

```
scmVersion {
    tag {
        prefix = release

        branchPrefix = [
            "^legacy/.*" : "legacy" // if none matched, prefix is taken, so only put the legacy pattern here
        ]
    }
}
```

szpak:
API looks good to me. It looks consistent with `branchVersionCreators`, so should be easy to memorize.

adamdubiel:
1.3.4 has been released :) It took me a while because i decided to do a major refactor along with this issue: revolutionary spirit. Should have released version with this feature before the refactor though.

szpak:
I've seen that you did some New Year cleanups in code. We will see how good regressions tests really are ;).

Btw, it would be probably good to also have (one day) some [nebula-test](https://github.com/nebula-plugins/nebula-test/)/TestKit acceptance tests for the main paths.

adamdubiel:
Yes, thats on my todo list as a next step (see description in cleanup PR) :) After cleaning up it all should be much easier to do.

PR #121, opened 2015-12-28
separating Gradle from core domain

adamdubiel:
When started working on #114 i got a bit fed up of *if* statements spreading through the code, followed by tight coupling with Gradle APIs. They were contained in factory methods, but still it became apparent, that core domain relies heavily on Gradle and it's impossible to use one without another, which has been pointed in #119 (and other conversations i had with @zygm0nt). This pull request is a major step towards separating core domain from Gradle.

Main change is the creation of **Properties** set of objects, which separate Gradle plugin configuration from properties that core services use to manage version. This is an ACL (anti corruption layer), that holds all mutable objects away from core domain and also makes testing easier (no need to rely on Gradle Project anymore).

Smaller change includes creation of simple and crude logging abstraction, which might allow on plugging other logging mechanisms without breaking Gradle-specific features (namely: `quiet`).

I tested the code using integration tests as well as using the sandbox project, it seems i didn't break anything :)

Next steps probably will include:

* separating core into other Gradle module
* writing Gradle plugin tests using [Gradle TestKit](https://docs.gradle.org/2.10/userguide/test_kit.html)

The properties and logging API is not stable, it might change during further separation process.


adamdubiel:
build fails with some ugly **buffer overflow detected** at `Java_java_net_Inet4AddressImpl_getLocalHostName`, i will take a look at it tomorrow, but seems to be Travis issue.

src/main/groovy/pl/allegro/tech/build/axion/release/ReleasePlugin.groovy
```diff
@@ -17,7 +17,7 @@ class ReleasePlugin implements Plugin<Project> {
     
     public static final String PUSH_RELEASE_TASK = 'pushRelease'
 
-    public static final String MARK_NEXT_VERSION_TASK = 'markNextVersionTask'
+    public static final String MARK_NEXT_VERSION_TASK = 'markNextVersion'
```
adamdubiel: this was a mistake, but thanks to Gradle guessing task names `markNextVersion` worked perfectly; since documentation holds `markNextVersion` and this is the only occurrence of `markNextVersionTask`, i assume changing it to the right value won't break any build

PR #125, opened 2016-01-08
Snapshot dependencies check added

### LINKED ISSUE #41: release plugin needs to validate that build is not using snapshot dependencies
If a build has included snapshot libraries, the release plugin should fail as it is not reproducible build script.

vbuell:
Closes #41. 
This adds snapshot dependencies check functionality.

### DIFF
--- a/src/main/groovy/pl/allegro/tech/build/axion/release/domain/SnapshotDependenciesChecker.groovy
+++ b/src/main/groovy/pl/allegro/tech/build/axion/release/domain/SnapshotDependenciesChecker.groovy
@@ -1,0 +1,27 @@
+package pl.allegro.tech.build.axion.release.domain
+
+import org.gradle.api.Project
+import org.gradle.api.artifacts.Dependency
+
+class SnapshotDependenciesChecker {
+
+    Collection<String> snapshotVersions(Project project) {
+        Collection<String> projectVersions = project.allprojects.collect {toFullVersion(it)}
+        Collection<String> allDependenciesVersions = project.allprojects.collect {
+            it.configurations.collect { config ->
+                config.allDependencies.findAll {isSnapshot(it)}.collect {toFullVersion(it)}
+
+            }
+        }.flatten().unique()
+        allDependenciesVersions.removeAll(projectVersions)
+        return allDependenciesVersions
+    }
+
+    boolean isSnapshot(Dependency dependency) {
+        dependency.version?.endsWith("-SNAPSHOT")
+    }
+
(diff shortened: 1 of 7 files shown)


vbuell:
And yeah. I remember that you were planning to use Pre-build hooks, but they doesn't help much here since HookContext doesn't have an access to project instance (which is for good reason).

src/main/groovy/pl/allegro/tech/build/axion/release/domain/SnapshotDependenciesChecker.groovy
```diff
@@ -0,0 +1,27 @@
...
+        allDependenciesVersions.removeAll(projectVersions)
+        return allDependenciesVersions
+    }
+
+    boolean isSnapshot(Dependency dependency) {
+        dependency.version?.endsWith("-SNAPSHOT")
+    }
+
+    String toFullVersion(it) {
+        "${it.group}:${it.name}:${it.version}"
```
szpak: I know that in case of Gradle in many places you cannot be sure what object is inside, but here `Project` should be quite safe. IMHO this code would be more readable with less `def`s.

adamdubiel:
I will give it some more look in the evening, but looks good to me, thanks! You are right, hooks are no good for things like this.

src/main/groovy/pl/allegro/tech/build/axion/release/domain/SnapshotDependenciesChecker.groovy
adamdubiel: hm, i see only 2 collection defs here, which look quite necessary, maybe give some groovy-magic sample?

szpak: I think that @vbuell rebased the code in the meantime (hmm, usually I was able to get "outdated diff", but in that case it seems to be not possible outside a small diff on the conversation page), from something like:
```
def projectVersions = ...
def allDependenciesVersions = ...
```
I also didn't like `String toFullVersion(it) {}` and it thought that `Project` could be used there, but it seems that there could be also a `Dependency` instance which has nothing common with `Project` (regarding a class hierarchy).

adamdubiel: Okay, i see :)

adamdubiel:
Thanks for another contribution @vbuell :)

PR #131, opened 2016-03-17
Add Java 6 compatibility

benfortuna:
Compile as Java 6 to support earlier JVMs

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -39,7 +39,7 @@ scmVersion {
 group = 'pl.allegro.tech.build'
 version = scmVersion.version
 
-sourceCompatibility = '1.7'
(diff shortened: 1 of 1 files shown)


adamdubiel:
Thanks! Is it really enough?

szpak:
@benfortuna Have you built it locally using Java 6?

benfortuna:
Hi,

I have just attempted building locally but it has problems. In short, the latest JGit has a min Java 7, dependency since 4.0. I tried downgrading to 3.7, but GrGit also has a min Java 7 requirement. Unfortunately this isn't really going to be possible without GrGit providing a Java 6-compatible version, so I'll close this PR.

regards,
ben

Issue #135, opened 2016-04-24
Transition from alpha to rc


himslm01:
I'm looking for a little help with workflow.

Lets say I've been developing a component and I have done a `markNextVersion` to create an alpha tag `release-1.0.0-alpha`.

When I think the component is ready I would deploy it to the Quality Assurance team. That sounds like a release candidate. That should be a `release-1.0.0-rc1` tag.

I already have `scmVersion.versionIncrementer 'incrementPrerelease'` in my build.gradle, but what would be the right way for me to have the -rc1 tag created?

I think I expected a `makeReleaseCandidate` task - but since there isn't one I must be missing something.
I've also failed to find any documentation on how `incrementPrerelease` was used in a workflow.

Could you point me in the right direction.

adamdubiel:
There is no task for this kind of thing. What you should do instead is run `./gradlew release -Prelease.verison=1.0.0-rc1`. This will start the `rc` count. Subsequent releases in your configuration will result in bumping the rc version (rc1, rc2...). When you are ready to go with full version, you should once again use release task with forced version: `./gradlew release -Prelease.verison=1.0.0`.

himslm01:
Thanks Adam.
Do you think there should be built-in tasks for those two actions?

adamdubiel:
I think maybe not a task itself, but some kind of version incrementation strategies that would do this, which could be set per release run:

```
./gradlew release -Prelease.versionIncrementer=alphaToRc
./gradlew release -Prelease.versionIncrementer=rcToStable
```

If i understand correctly you would like to have all those steps automated and passing some forced version is not so convenient, as it requires human intervention and version parsing.

himslm01:
I see where you're going.

Obviously, what I'm after is something that requires the least thought process and is the simplest to automate in a CI process.

Using your suggestion, following a usual development flow, we'd start 0.1.0 and `release` some snapshots
```
0.1.0-snapshot
0.1.1-snapshot
0.1.2-snapshot
```
We've finished most coding and convert to alpha calling the `markNextVersion` task with a `release.nextVersion` of `1.0.0`.
```
1.0.0-snapshot
```
And then convert alpha to rc calling the `release` task with a specific `versionIncrementer`
```
1.0.0-rc1
```
`release` a few of those
```
1.0.0-rc2
1.0.0-rc3
```
When we're happy, `release` with a specific `versionIncrementer` to create the 1.0.0
```
1.0.0
```

[...]

Issue #132, opened 2016-03-22
Three commits with new Release version


OlgaMaciaszek:
We've been using this plugin for quite a long time in the Accurest project, but after switching to version 1.3.4, we have the following issue:

after running `./gradlew createRelease` three commits with a new relesase version get created in place of one. This is what is returned by `git log`:

```
commit c36f7b0cf25efdb2bb6b19475612f51a6e8bb449
Date:   Tue Mar 22 08:46:49 2016 +0100
Release version: 1.0.1    
[ci skip]

commit 4d2d3c6f9723bc0dba92afa9e1a9870cfb898e76
Date:   Tue Mar 22 08:46:49 2016 +0100
Release version: 1.0.1    
[ci skip]

commit ab8a5b03fcf3cf999530e75b36e7c36394a283bb
Date:   Tue Mar 22 08:46:49 2016 +0100
Release version: 1.0.1
[ci skip]
```

This issue is repeatable.

adamdubiel:
Hi, i just tried to reproduce this in both single module and multi-module project ([hermes](https://github.com/allegro/hermes) with updated axion-release version) and i failed to see the issue you described.

My guess is you have some kind of multi-module project and `createRelease` task gets called for each module, but posting relevant parts of configuration would help :)

adamdubiel:
I downloaded [Accurest](https://github.com/Codearte/accurest) sources and i see the issue now.

First i run `./gradlew release -Prelease.local` after changing gradle wrapper version to `2.11`, which i have locally. It produced only one commit. It seemed weird, so i grit my teeth and downloaded gradle `2.4` on mobile internet to see that in fact 3 commit messages showed up... looks like some issue with earlier versions of Gradle - i can investigate the exact version once i have broadband internet connection, for now i would propose to change wrapper version.

OlgaMaciaszek:
Ok. Will raise it. Thanks @adamdubiel

OlgaMaciaszek:
After Gradle in AccuREST has been raised to 2.12, we still have this bug.

szpak:
It seems to be not related to multiproject build. I had the same in https://github.com/szpak/spock-global-unroll (in the remote repo there is just one commit as I rebased it locally before push).

szpak:
[...]

PR #141, opened 2016-05-25
Tag selector

### LINKED ISSUE #139: latestTag is the last rc tag instead of the non-rc tag
When there are tags

1.0.0
1.0.0-rc1
1.0.0-rc2

the plugin show current version as `1.0.0-rc2` and find that as the latestTag. It should find the tag without the prerelease suffix.

levsa:
Added support for a "tagSelector" closure in tag configuration, which is called to select among multiple tags on the same commit. Defaults back to selecting the last tag in the list of tags returned from scm (same behaviour as before). This allows to resolve #139 (see also discussion in #135).

### DIFF
--- a/src/main/groovy/pl/allegro/tech/build/axion/release/infrastructure/git/GitRepository.groovy
+++ b/src/main/groovy/pl/allegro/tech/build/axion/release/infrastructure/git/GitRepository.groovy
@@ -32,9 +32,9 @@ class GitRepository implements ScmRepository {
     private static final String GIT_TAG_PREFIX = 'refs/tags/'
 
     private final TransportConfigFactory transportConfigFactory = new TransportConfigFactory()
-    
+
     private final File repositoryDir
-    
+
     private final Grgit repository
 
     private final ScmProperties properties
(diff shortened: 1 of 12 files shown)


src/main/groovy/pl/allegro/tech/build/axion/release/infrastructure/git/GitRepository.groovy
```diff
@@ -159,38 +159,50 @@ class GitRepository implements ScmRepository {
...
             return ScmPosition.defaultPosition()
         }
 
         Map tags = repository.tag.list()
                 .grep({ def tag = it.fullName.substring(GIT_TAG_PREFIX.length()); tag ==~ pattern && !(tag ==~ inversePattern) })
-                .inject([:], { map, entry -> map[entry.commit.id] = entry.fullName.substring(GIT_TAG_PREFIX.length()); return map; })
+                .inject([:], { map, entry ->
+                    if (map[entry.commit.id] == null) {
+                        map[entry.commit.id] = [] as List<String>
+                    }
```
vbuell: Dealing with non existing values of map can be simplified by using [:].withDefault {p -> []}

levsa: I looked for something similar. But that would break the null-check later on, which should check for empty list instead (i.e. just do if (map[id]) return )

adamdubiel:
Great stuff, thanks! Could you squash the commits before merge?

levsa:
I think it can be done during merge?

https://github.com/blog/2141-squash-your-commits

I'm not sure if I can do it now that the commits are pushed already?

adamdubiel:
Good call, didn't have a chance to try out this new Github Merging stuff :) Thanks once again.

levsa:
Great, thank you!

adamdubiel:
BTW i never release the version - now it's here: 1.3.5. Sorry it took so long :)

dafelipe130:
Total groovy/gradle newbie here. How do you go about telling the tagSelector option to go for the default setting which is to take the tag without the 'rc' identifier?

adamdubiel:
@dafelipe130 what do you mean by this? There are usually no `-rc` tags.. can you show the structure?

dafelipe130:
Basically, I'm facing the same situation described here #139 
In a single commit, I have both a release candidate (app-1.0.0-rc.1) and release tag (app-1.0.0). Running `currentVersion` with that commit as my HEAD picks up the release candidate tag instead of the release tag which is different behavior when compared to using git describe --tags --abbrev=0

dafelipe130:
I fiddled around a bit on this one and have settled with just configuring the tagSelector to always get the first item in the _tags_ object
`tagSelector = { tags -> return tags[0] }`
... which works fine with our very specific use case

However, this doesn't respect the next version marker tag if its situated in the same commit as a release/pre-release tag. The next version marker tag is not handled differently anymore and is treated as any other tag contrary to what is specified [here](http://axion-release-plugin.readthedocs.io/en/latest/configuration/next_version.html).

This results in a scenario wherein axion reports the current version as the incremented version of the immediate release tag instead of the version specified by the next version marker.

I suppose I can add a condition to check if the last tagged commit has a next version marker then abstain the plugin from using the tagSelector option to honor said marker. Or just instruct users to first commit something then apply the marker.

Please feel free to chime in if I'm missing anything here.

adamdubiel:
After analysing what you said I think bringing in `tagSelector` created a bit of a mess to already messy `-alpha` logic. What i think is that this *nextVersionMarker* section needs some more thinking and refactoring to make sense - i had a lot of signals about misbehaviour recently (on github and inside the company).

adamdubiel:
I think i figured out some basic rules of better implementation. Current messy implementation stems from not wanting to interact with Git more than absolutely necessary, but i think that it only made the code hacky aka *clever* thus unmaintainable in the long run. Stay tuned :)

adamdubiel:
So after some time i finally managed to refactor all the code that will make the "many tags" problem go away. The tests are green, but i want to add more integration tests to the mix. The branch is here: https://github.com/allegro/axion-release-plugin/tree/fix-tag-ordering

The main highlight is reading all tags on commit and choosing the one with the highest version. Currently there is no way to customise the behaviour and i'm not really sure it's needed - should be no problem to add it though.

After finishing work it will be released as 1.5.0, due to small API change in two places which are probably used rather rarely - custom deserializers.

szpak:
By `"many tags" problem` you mean #132?

PR #142, opened 2016-05-26
Rc incrementors

levsa:
Created predefined incrementers:
* createMajorRC: Create a prerelase with -RC1 for the next major version
* createMinorRC: Create a prerelease with -RC2 for the next minor version
* incrementPrereleaseOrMinor: Increment prerelease if already on prerelease, otherwise minor

Also created "createFinal" to release a final version from a prerelease, but couldn't make it work without using "release.forceSnapshot". There is an ignored test for it. It can be used as follows though:

```gradle
task releaseFinal {
    finalizedBy release
    doFirst {
        project.ext["release.versionIncrementer"] = 'createFinal'
        project.ext["release.forceSnapshot"] = true
    }
}
```

### DIFF
--- a/src/main/groovy/pl/allegro/tech/build/axion/release/domain/PredefinedVersionIncrementer.groovy
+++ b/src/main/groovy/pl/allegro/tech/build/axion/release/domain/PredefinedVersionIncrementer.groovy
@@ -65,8 +65,8 @@ enum PredefinedVersionIncrementer {
     CREATE_FINAL('createFinal', { VersionIncrementerContext context, Map config ->
         throwErrorIfNotPrerelease(context)
         return new Version.Builder()
-                .setNormalVersion(context.currentVersion.normalVersion)
-                .build();
+            .setNormalVersion(context.currentVersion.normalVersion)
+            .build();
     }),
 
     BRANCH_SPECIFIC('branchSpecific', { VersionIncrementerContext context, Map config ->
--- a/src/main/groovy/pl/allegro/tech/build/axion/release/domain/PredefinedVersionIncrementer.groovy
+++ b/src/main/groovy/pl/allegro/tech/build/axion/release/domain/PredefinedVersionIncrementer.groovy
@@ -46,6 +46,29 @@ enum PredefinedVersionIncrementer {
         return context.currentVersion.incrementPatchVersion()
     }),
 
+    CREATE_MAJOR_RC('createMajorRC', { VersionIncrementerContext context, Map config ->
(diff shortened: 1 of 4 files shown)


szpak:
Similar approach could be also useful with beta versions. I wonder is it worth to extend it (probably in an another PR) to support parameterized incrementers (with prefix as a parameter like `beta1`, `beta.1`, `rc-1`, etc.) or it would be enough to just have two additional tasks `CREATE_MAJOR_BETA` and `CREATE_MINOR_BETA`?

adamdubiel:
I wonder if we could use some Gradle magic here and extract the suffix from task name? What i mean is dynamic tasks like `create[Minor/Major][Suffix]`.

levsa:
The suffix could be configured in the scmVersion configuration like other configuration, e.g. in the versionIncrementer section perhaps?

Alternatively, a project property, "release.releaseCandidateSuffix" could be used. Then it can be set from the task as you said. Task rules can be written like so:

```gradle
tasks.addRule("Pattern: create<Major|Minor|Final><Suffix> \t- create a release candidate") { String taskName ->
    if (taskName.startsWith("create")) {
        logger.info "Adding $taskName"
        def p = /create(Major|Minor|Final)(.*)/
        def majorMinorFinal = taskName.replaceAll(p) { all, mm, s -> mm }
        def suffix = taskName.replaceAll(p) { all, mm, s -> s }
        task(taskName) {
            group = 'Creating release candidates'
            description = "Creates a ${majorMinorFinal} with suffix ${suffix}"
            doFirst {
                logger.lifecycle "Create ${majorMinorFinal} with suffix: ${suffix}"
                project.ext["release.releaseCandidateSuffix"] = suffix
                if (majorMinorFinal == 'Final') {
                    project.ext["release.versionIncrementer"] = 'createFinal'
                    project.ext["release.forceSnapshot"] = true
                } else {
                    project.ext["release.versionIncrementer"] = "create${majorMinorFinal}RC"
                }
            }
        }
    }
}
```

But I think it was much short to write the separate tasks createMajorRC, createMinorRC and createFinal.

adamdubiel:
Could you give me a day or two to try other implementation? If i fail to find time/implement we will go with this pr.

levsa:
Of course, no hurries!

jakubkrolewski:
I was just about to request this feature, when a noticed this PR. Great work :) Please, remember to update the docs when you decide how the tasks should look like.

bgalek:
we can get back to it if needed

PR #152, opened 2016-08-18
Update changelog.rst

vi4m:
Actually, I love reading changelogs during my spare time ;-)


adamdubiel:
Good to have proper proof-reading, even if post-publishing ;)

Issue #146, opened 2016-08-05
Tag selector picks rc over release version


asarkar:
In [ScmRepository.groovy](https://github.com/allegro/axion-release-plugin/blob/1de6a06b4c1ff88291e039223820cc5a7c80b096/src/main/groovy/pl/allegro/tech/build/axion/release/domain/scm/ScmRepository.groovy)

```
public static final Closure<String> LAST_TAG_SELECTOR = { List<String> tags ->
    tags && tags.size() > 0 ? tags[-1] : null
}
```

The tag selector picks the last one in the list but the order of tags seems to be undeterministic. This causes `2.1.0-rc1` to be picked over `2.0.0` from a list `[2.0.0, 2.1.0-rc1]`.
Why not do this `tags.split { ~/.*-rc\d+$/ }.find().max()`?

adamdubiel:
Go ahead :) In my opinion having multiple tags on a single commit is kind of a corner case. However if you wanted to change the default behaviour, then i'm afraid this regexp won't suffice - am i right it would not work for `[2.0.0, 2.1.0]`? Ideally it would just pick the highest version, but to do that it needs to be parsed.

If you have time to create this kind of pull request it would be a welcome addition to axion-release :)

abhijitsarkar:
@adamdubiel 
> it would not work for [2.0.0, 2.1.0]

Why not?

```
def tags = ['2.0.0', '2.1.0']
tags.split { ~/.*-rc\d+$/ }.find().max()
Result: 2.1.0
```

adamdubiel:
Hm, you are right - it works nicely. Okay, great - how about a PR with this small change? :)

abhijitsarkar:
A quick PR attempt failed 18 tests. That's more than I can handle right now. Perhaps another day.

Issue #149, opened 2016-08-12
Version is not resolved correctly when alpha tag and version tag point to the same commit


Endron:
I run into a bug the other day. While preparing to release a new version of our application I added alpha tags to SCM to set the versions I wanted. I did this as shown in the documentation by executing 

```
./gradlew markNextVersion -Prelease.version=<next.version>
```

This created the tag as expected.
I then created a new release by running

```
./gradlew release -Prelease.disableUncommittedCheck -Prelease.disableRemoteCheck
```

This also tagged the version as expected.
Now we have an alpha tag and an release tag pointing to the same commit in the repository,

But then it went downhill when I tried to have the plugin resolve the version 

```
./gradlew currentVersion
```

Returned a SNAPSHOT version. As the version was what I initially set using "markNextVersion" it looks like the release tag was ignored. The problem with this is that now instead of building the release version I want to deploy to our repository manager (Artifactory) I am still building a SNAPSHOT version.

**Environment**
- Plugin version used was 1.3.4
- "markNextVersion" was executed on Windows 7 running with Java 8 and Gradle 2.12
- "release" was executed on a Linux CI server running Bamboo. Also Java 8 and Gradle 2.12
- the prefix for our release tags was "usl-nominator" and "batch-uploader" (I had this problem with two different applications in different git repos.)

**Workaround**
This does not happen when the alpha and the release tag do not point to the same git commit. So if you put some dummy commit between "markNextVersion" and "release" you are save.

**TL;DR**
If an alpha tag and an release tag point to the same git commit the version is not recognized correctly. The version is returned as SNAPSHOT when it should be the release version.

adamdubiel:
Hi, this is the same issue as discussed in comments in #141 - i will be working on solution to this.

adamdubiel:
Closed in 1.5.0 by #156

radimk:
Note that #156 didn't fix it. Fortunatelly it is fixed not by https://github.com/allegro/axion-release-plugin/commit/634ff11e1c414d2df03e3cfb0587618b78e527e5 in 1.7.2 (which BTW is only available from mavenCentral and not from plugins.gradle.org repo as described in #197).

PR #156, opened 2016-09-18
refactor tag retrieval and add tag ordering by version

adamdubiel:
Fixes: #146 #149 

Effect of this pull request is adding ordering of tags when there are more than one on a commit. Tags are ordered by highest version. This should fix all problems with multiple tags.

As a side effect, code retrieving tags has been refactored and should be easier to understand and maintain in the future.

Also `tag selector` feature introduced in #141 has been removed - @levsa please let me know if it should be reimplemented, or is current implementation good enough.


adamdubiel:
@szpak continuing #141 here: no, i meant having multiple tags on one commit

src/main/groovy/pl/allegro/tech/build/axion/release/domain/VersionResolver.groovy
```diff
@@ -6,50 +6,99 @@ import pl.allegro.tech.build.axion.release.domain.properties.TagProperties
...
+                version      : version,
+                isNextVersion: isVersionNextVersion.get(version),
+                noTagsFound  : versions.isEmpty()
         ]
     }
 
+    private static class VersionComparator implements Comparator<Version> {
+        @Override
+        int compare(Version o1, Version o2) {
+            return -1 * o1.compareTo(o2)
```
vbuell: I believe you may use Collections.reverseOrder() instead.

adamdubiel: right :)

Endron:
Thank you!

PR #163, opened 2016-11-16
Fix typo: on[]y -> on[l]y

lgawin:
Fix typo: on[]y -> on[l]y

PR #165, opened 2017-01-03
Revert "Fix typo: on[]y -> on[l]y"

Foxytine31:
Reverts allegro/axion-release-plugin#163


adamdubiel:
I don't see why should i revert this typo fix.. was it a misclick?

PR #176, opened 2017-02-08
replaced deprecated release.nextVersion with release.version

piotrchowaniec:
In the documentation and examples there is release.nextVersion property used. As it is deprecated now, should be replaced with release.version.


adamdubiel:
I was just about to create PR by myself, thanks for the fix :)

piotrchowaniec:
Nice timing :)

PR #178, opened 2017-03-02
Return the greatest version found in a RevWalk, not just the most recent found in the list

ProTrent:
This compares all of the versions returned in a RevWalk and returns the greatest version found. Helps to fix #52.

This is just a start and works for me for now; however, it should give you a general idea of what we need.

PR #179, opened 2017-03-08
Add EditorConfig configuration

szpak:
To address @vbuell's comment in the recent [PR](https://github.com/allegro/axion-release-plugin/pull/178#pullrequestreview-25897851).

We had similar problem with formatting in PRs in our projects and [EditorConfig](http://editorconfig.org/) helped in simple scenarios (such as spaces vs tabs). It is supported out-of-box by Idea and some other IDEs/editors with support in several more available as plugins.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+# EditorConfig: http://EditorConfig.org
+root = true
+
+[*]
+indent_style = space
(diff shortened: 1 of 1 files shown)


adamdubiel:
Thanks, lets give it a try :)

PR #193, opened 2017-07-18
Adding optional VersionConfig to each task

maria-camenzuli:
At the moment, the tasks in the plugin cannot be reused without applying the entire plugin, since they rely on the `VersionConfig` extension.

In this pull request, I add an `Optional` `VersionConfig` input to each task so that other plugins can use a single task, or a subset of them, and set the `VersionConfig` directly for each task.

What I'm trying to enable here is something like the following.
```groovy
def exampleTask = project.tasks.create('exampleTask', ReleaseTask)
exampleTask.group = 'example'
exampleTask.description = 'Does example things...'
exampleTask.configure {
    versionConfig = new VersionConfig(project)
    versionConfig.versionIncrementer 'incrementMajor'
    // .. etc
}
exampleTask.doFirst {
    // .. stuff
}
```

### DIFF
--- a/src/main/groovy/pl/allegro/tech/build/axion/release/CreateReleaseTask.groovy
+++ b/src/main/groovy/pl/allegro/tech/build/axion/release/CreateReleaseTask.groovy
@@ -1,18 +1,23 @@
 package pl.allegro.tech.build.axion.release
 
 import org.gradle.api.DefaultTask
+import org.gradle.api.tasks.Optional
 import org.gradle.api.tasks.TaskAction
 import pl.allegro.tech.build.axion.release.domain.Releaser
+import pl.allegro.tech.build.axion.release.domain.VersionConfig
 import pl.allegro.tech.build.axion.release.infrastructure.di.Context
 import pl.allegro.tech.build.axion.release.infrastructure.di.GradleAwareContext
 
 class CreateReleaseTask extends DefaultTask {
 
+    @Optional
+    VersionConfig versionConfig
+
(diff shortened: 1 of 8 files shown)


adamdubiel:
Thanks for the pull request! However i'm on vacation for the next two
weeks, so i will review it after coming back. Thanks for your patience :)

maria-camenzuli:
Not a problem. Enjoy your vacation!

maria-camenzuli:
Not entirely sure this is useful for all tasks to be honest. I did this in a rush because I had a situation where I wanted to reuse the `ReleaseTask` specifically without applying the entire plugin. Your call 😃

adamdubiel:
All in all it looks good for me. Could you maybe describe your use case so i could add it as an example in documentation?

maria-camenzuli:
In my use case, I wanted to write my own custom task that performs a release, and then follows that up with some other logic. So I wanted to create a new task of type `ReleaseTask`, and then add some logic to be executed with it. Since I only needed the `ReleaseTask`, I wanted to avoid applying the entire plugin, since this would pollute my build with tasks that I didn't need.

adamdubiel:
I will merge it and add some integration tests :)

adamdubiel:
Build is failing because of JDK 7 problems with Gradle plugin portal TLS.

adamdubiel:
I released 1.8.0 (and 1.8.1) with this feature :) Once again thanks for the contribution!

PR #201, opened 2017-10-06
Customization test

adamdubiel:
test for task customizations


adamdubiel:
@maria-camenzuli i added tests which uncovered a problem with not using proper `VersionConfig` object. After merging this, i will release `1.8.0`.

Issue #203, opened 2017-10-20
Push task silently ignores rejection from remote server


adamdubiel:
If there is any kind of remote-side validation of pushes and it fails `axion-release` silently ignores errors and prints success, instead of failing the build.

#### Steps to reproduce

Create new repo:

```
mkdir restricted-remote-repo
git init
touch some_file
git add -A
git commit -m "Initial commit"
```

Add pre-receive hook which always fails with error message:

```
echo "echo 'This push will fail!'; exit 1" >  .git/hooks/pre-receive
chmod +x  .git/hooks/pre-receive
```

Clone newly created repo into new dir:

```
git clone /home/..../restricted-remote-repo axion-release-repo
```

Add build gradle with axion-release and run release command:

```
gradle release -Prelease.disableChecks
```

This should end with success, while under the hood everything failed and no commits have been pushed to remote. However tag has already been created, so we end up with bad state.

#### Expected behaviour

Release task should fail, printing message from remote. If task failed, created tag should be removed (tag and push should be in a sort of "transaction").

#### Implementation notes

* most of changes can be done in `GitRepository` class
* `GrGit#push` does not even return error status (nor throws exceptions), so raw `JGit` has to be used
* for testing: when used with local repo, `JGit` will not run `pre-receive` hooks (no support for it), so some other fail has to be simulated for testing purposes (maybe ahead of remote?)

mareckmareck:
I will check this out and attempt to come up with something

PR #204, opened 2017-10-21
Fixes silently ignoring push failure while leaving created tag intact (#203)

mareckmareck:
During release task a new tag is created and then a push is attempted. However, if said push fails due to remote-side validation, the task still ends as a success while leaving the created tag. This creates the inconsistent state.

After applying this fix the task will fail with exception received from the remote during push and the tag will be removed (if it was created). This solves the issue #203

### DIFF
--- a/src/main/groovy/pl/allegro/tech/build/axion/release/domain/Releaser.groovy
+++ b/src/main/groovy/pl/allegro/tech/build/axion/release/domain/Releaser.groovy
@@ -4,6 +4,7 @@ import com.github.zafarkhaja.semver.Version
 import pl.allegro.tech.build.axion.release.domain.hooks.ReleaseHooksRunner
 import pl.allegro.tech.build.axion.release.domain.logging.ReleaseLogger
 import pl.allegro.tech.build.axion.release.domain.properties.Properties
+import pl.allegro.tech.build.axion.release.domain.scm.ScmException
 import pl.allegro.tech.build.axion.release.domain.scm.ScmService
 
 class Releaser {
@@ -22,7 +23,7 @@ class Releaser {
         this.hooksRunner = hooksRunner
     }
 
-    void release(Properties rules) {
+    Optional<String> release(Properties rules) {
         VersionContext versionContext = versionService.currentVersion(rules.version, rules.tag, rules.nextVersion)
         Version version = versionContext.version
 
@@ -35,12 +36,31 @@ class Releaser {
             repository.tag(tagName)
 
             hooksRunner.runPostReleaseHooks(rules.hooks, rules, versionContext, version)
+            return Optional.of(tagName)
         } else {
-            logger.quiet("Working on released version ${version}, nothing to release.")
+            logger.quiet("Working on released version ${version}, nothing to release")
+            return Optional.empty()
+        }
+    }
+
+    void releaseAndPush(Properties rules) {
+        Optional<String> releasedTagName = release(rules)
+
+        try {
+            pushRelease()
+        } catch (ScmException e) {
+            releasedTagName.ifPresent { rollbackRelease(it) }
+            throw e
         }
     }
 
     void pushRelease() {
         repository.push()
+    }
+
+    private void rollbackRelease(String tagName) {
+        logger.quiet("Removing tag: $tagName")
+        repository.dropTag(tagName)
+        logger.quiet("Tag $tagName removed")
     }
 }
--- a/src/main/groovy/pl/allegro/tech/build/axion/release/domain/scm/ScmService.groovy
+++ b/src/main/groovy/pl/allegro/tech/build/axion/release/domain/scm/ScmService.groovy
@@ -23,12 +23,27 @@ class ScmService {
         repository.tag(tagName)
     }
 
+    void dropTag(String tagName) {
+        try {
+            repository.dropTag(tagName)
+        } catch (ScmException e) {
+            logger.quiet("Exception occurred during removing the tag: ${e.message}")
+            throw e
+        }
+    }
+
     void push() {
-        if (!localOnlyResolver.localOnly(this.remoteAttached())) {
+        if (localOnlyResolver.localOnly(this.remoteAttached())) {
+            logger.quiet("Changes made to local repository only")
+            return
+        }
+
+        try {
             logger.quiet("Pushing all to remote: ${scmProperties.remote}")
             repository.push(scmProperties.identity, scmProperties.pushOptions())
-        } else {
-            logger.quiet("Changes made to local repository only")
+        } catch (ScmException e) {
+            logger.quiet("Exception occurred during push: ${e.message}")
+            throw e
         }
     }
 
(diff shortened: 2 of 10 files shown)


src/main/groovy/pl/allegro/tech/build/axion/release/domain/Releaser.groovy
```diff
@@ -1,9 +1,11 @@
 package pl.allegro.tech.build.axion.release.domain
 
 import com.github.zafarkhaja.semver.Version
+import org.eclipse.jgit.api.errors.GitAPIException
```
adamdubiel: releaser is part of domain, while `jgit` is an implementation detail - this exception should be wrapped into some kind of domain exception (like `ScmException`). `jgit` should be only visible within `GitRepository`

src/main/groovy/pl/allegro/tech/build/axion/release/domain/Releaser.groovy
```diff
@@ -35,12 +37,33 @@ class Releaser {
...
+    void releaseAndPush(Properties rules) {
+        def releasedTagName = release(rules)
+
+        try {
+            pushRelease()
+        } catch (ScmPushFailedException | GitAPIException e) {
+            logger.quiet("Exception occurred during pushing the release: ${e.message}.")
+            releasedTagName.ifPresent{ tagName ->
+                logger.quiet("Removing tag: $tagName.")
+                rollbackRelease(tagName)
```
adamdubiel: i would add "tag $tagName removed" and wrap rollbackRelease in try/catch, so if any error occurs in between, we know that the tag has not been removed

src/main/groovy/pl/allegro/tech/build/axion/release/domain/scm/ScmService.groovy
```diff
@@ -23,12 +23,16 @@ class ScmService {
...
     }
 
+    void dropTag(String tagName) {
+        repository.dropTag(tagName)
+    }
+
     void push() {
         if (!localOnlyResolver.localOnly(this.remoteAttached())) {
-            logger.quiet("Pushing all to remote: ${scmProperties.remote}")
+            logger.quiet("Pushing all to remote: ${scmProperties.remote}.")
```
adamdubiel: i don't like dots ;)

src/main/groovy/pl/allegro/tech/build/axion/release/infrastructure/git/GitRepository.groovy
```diff
@@ -90,26 +99,36 @@ class GitRepository implements ScmRepository {
...
     void push(ScmIdentity identity, ScmPushOptions pushOptions, boolean all) {
-        identity.useDefault ? callPush(pushOptions, all) : callLowLevelPush(identity, pushOptions, all)
+        callPush(identity, pushOptions, all)
     }
 
-    private void callPush(ScmPushOptions pushOptions, boolean all) {
+    private void callPush(ScmIdentity identity, ScmPushOptions pushOptions, boolean all) {
         if (!pushOptions.pushTagsOnly) {
-            repository.push(remote: pushOptions.remote, all: all)
+            def regularPush = pushCommand(identity, pushOptions.remote, all)
```
adamdubiel: the only difference between `pushCommand` here and next invocation is `setPushTags`, so how about:

```
def pushCommand = pushCommand(...)
if(pushOptions.pushTagsOnly) {
    pushCommand = pushCommand.setPushTags()
}
```

Also please use types instead of `def`, as i do in other places

src/main/groovy/pl/allegro/tech/build/axion/release/infrastructure/git/GitRepository.groovy
```diff
@@ -90,26 +99,36 @@ class GitRepository implements ScmRepository {
...
-        if (!pushOptions.pushTagsOnly) {
-            pushCommand(identity, pushOptions.remote, all).call()
-        }
-        pushCommand(identity, pushOptions.remote, all).setPushTags().call()
+    private void verifyPushResults(Iterable<PushResult> pushResults) {
+        PushResult pushResult = ++pushResults.iterator()
+        Iterator<RemoteRefUpdate> remoteUpdates = pushResult.getRemoteUpdates().iterator()
+
+        remoteUpdates
+            .find { it.getStatus() == RemoteRefUpdate.Status.REJECTED_OTHER_REASON }
```
adamdubiel: i would go for `it.getStatus != SUCCESS`, so we log any errors (including rejection because of ahead of remote etc)

src/main/groovy/pl/allegro/tech/build/axion/release/domain/Releaser.groovy
```diff
@@ -35,12 +37,33 @@ class Releaser {
...
+            }
+            throw e
         }
     }
 
     void pushRelease() {
         repository.push()
```
adamdubiel: exception handling logic is missing here - this method is called from `PushRelease` task and displayed output should be the same (only without rollback)

src/main/groovy/pl/allegro/tech/build/axion/release/domain/Releaser.groovy
mareckmareck: That's so very true, silly of me to have missed this leak.

src/main/groovy/pl/allegro/tech/build/axion/release/domain/Releaser.groovy
mareckmareck: I actually stopped myself from making too many changes to existing structures, because that would start an endless stream of "Oh and while I'm at it, I'll change this this and that too" and possibly ruin some design.

That being said, it sounds very reasonable to apply what you suggest here since it would be really awkward for those two methods to act differently in case of an error.

src/main/groovy/pl/allegro/tech/build/axion/release/domain/scm/ScmService.groovy
mareckmareck: My old habits die hard

src/main/groovy/pl/allegro/tech/build/axion/release/infrastructure/git/GitRepository.groovy
mareckmareck: Since this comment is also true for situation that been before, I figured you had some reason to leave it that way. Happy to change it though.

src/main/groovy/pl/allegro/tech/build/axion/release/infrastructure/git/GitRepository.groovy
mareckmareck: I've also considered this, but there are some problems with this approach. There is a RemoteRefUpdate.Status.OK that we could use in place of suggested "SUCCESS", but:

1. UP_TO_DATE status is not really an error, so us failing on it doesn't sound right.
2. I think that applying the "! = OK" we actually lose a bit of control about what we are actually checking here. Even if we add UP_TO_DATE, in rather unlikely scenario when some new version of the library actually adds new "non-negative" status here, that check could lead to a lot of frustration.

But I find your suggestion rather pragmatic so I'll just go on and do it

PR #210, opened 2017-12-25
Tests for remote push rejection

adamdubiel:
Docker-based tests for remote push rejection. Due to the deficiencies of jgit hooks implementation (lack of pre-receive support), those tests could not be emulated in java.

Tests which require docker are isolated in `remoteTest` target. They require running docker demon. Creation of image and running the container are automated. `remoteTest` is also part of `check` target.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: false
+sudo: required
 
 language: java
 
@@ -8,3 +8,6 @@ after_success: ./gradlew jacocoTestReport coveralls
 
 jdk:
   - oraclejdk8
+
+services:
+  - docker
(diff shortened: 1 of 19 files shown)


adamdubiel:
@mareckmareck this tests your recent changes.

mareckmareck:
Impressive! 

I was considering similar approach, but at the time it felt rather heavy to start a VM/Docker container with SSH just to test one (minor) functionality, so I dropped this approach. Additionally it introduces new dependencies and increases the complexity of the build script in a nasty way.

On the other hand it works really smoothly once setup and now we know that my proposed change actually works :). Thanks for filling in the missing piece there.

adamdubiel:
I put it in separate configuration, so it will be mostly run by travis. Even though it is just one test now, i think i might add some more. SSH and interacting with remotes was mostly tested manually back when i was trying to figure out how it all works.

PR #213, opened 2018-01-03
#212 fix incorrect version after release when using markNextVersion a…

### LINKED ISSUE #212: useHighestVersion and markNextVersion don't work correctly together
The resolved version is incorrect after doing a release based on a next version marker when using useHighestVersion. It uses the next snapshot version instead of the current release version, making it impossible to publish the release with the correct version number.

Reproduction steps:
```
# ./gradlew cV
0.1.1-SNAPSHOT
# ./gradlew markNextVersion -Prelease.version=1.0.0
# git commit -m 'some work' --allow-empty
# ./gradlew release
# ./gradlew cV
1.0.1-SNAPSHOT // should be 1.0.0
```
The git log looks like this at this point:
```
cad6ee7 (tag: release-1.0.0) some work
d2c5794 (tag: release-1.0.0-alpha) set useHighestVersion to true
0567f21 (tag: release-0.1.0) initial commit with axion-release 1.8.1
```

cliffred:
Fixes #212

### DIFF
--- a/src/main/groovy/pl/allegro/tech/build/axion/release/domain/VersionResolver.groovy
+++ b/src/main/groovy/pl/allegro/tech/build/axion/release/domain/VersionResolver.groovy
@@ -108,7 +108,7 @@ class VersionResolver {
                                          boolean ignoreNextVersionTags,
                                          Pattern nextVersionTagPattern,
                                          VersionFactory versionFactory) {
-      List<Version> versions = []
+      Set<Version> versions = []
       Map<Version, Boolean> isVersionNextVersion = [:]
       Map<Version, TagsOnCommit> versionToCommit = new LinkedHashMap<>()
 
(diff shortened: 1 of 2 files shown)


adamdubiel:
Thanks for the fix!

I analysed the problem just to understand it better. From what i see if there are two commits which resolve to the same version (as in 1.1.0 and 1.1.0-alpha) commit metadata is taken from earlier one. In this case it is -alpha. Commit metadata, namely `isHead` property is crucial when determining if it is a snapshot or release. Taking `isHead` from wrong commit results in this bug.

cliffred:
That's exactly right. The fix just makes sure that whenever a final and alpha exist for the same version, the final is always regarded as newer.

adamdubiel:
Merged :) Plus i added a commit with short comment on this logic.

Would you like me to release this fix asap, or can i take my time to fix the other problem you reported (push fails with auth problems)?

cliffred:
Thanks :)
It can wait. We can work around this issue by deleting the alpha tag.

PR #216, opened 2018-01-24
Another jgit grgit update

bigwaff:
Updated jgit and grgit to latest versions. Build succeeds and unit tests pass. Of note -
- Latest grgit requires Java 8 (perhaps this is undesirable?)
- GitRepositoryTest used a grgit exception class removed in latest version; replaced with Exception

The issue which prompted this PR
- GitRepository.checkAheadOfRemote() does not have a unit test and failed in real life usage due to an api change in grgit BranchService. I'm not sure if there are other similar issues of this type lurking in the code base.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,7 +3,7 @@ plugins {
     id 'java-gradle-plugin'
     id 'maven'
     id 'jacoco'
-    id 'pl.allegro.tech.build.axion-release' version '1.8.1'
+    id 'pl.allegro.tech.build.axion-release' version '1.8.2'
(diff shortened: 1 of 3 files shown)


bigwaff:
I was thinking some more about the situation where clients may want to use different versions of libraries axion is dependent upon. One possible approach would be to use shadow plugin to create axion uber jar containing relocated versions of dependencies jgit, grgit, etc. This would insulate both clients and axion from classpath pollution.

adamdubiel:
Since i released 1.9.0 which does not use GrGit, i'm closing this PR.

Issue #215, opened 2018-01-23
Incompatibility with latest version of grgit plugin


bigwaff:
We use latest version of grgit plugin in our builds. We like to use axion but when both the latest version grgit plugin and latest version axion plugin are both declared plugins for a project, axion has following issue performing "release" task (possibly other tasks as well).
```
$ ./gradlew release --stacktrace

> Task :verifyRelease FAILED
Looking for uncommitted changes..

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':verifyRelease'.
> No such property: branch for class: org.ajoberstar.grgit.operation.BranchStatusOp

* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':verifyRelease'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeActions(ExecuteActionsTaskExecuter.java:100)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:70)
        at org.gradle.api.internal.tasks.execution.OutputDirectoryCreatingTaskExecuter.execute(OutputDirectoryCreatingTaskExecuter.java:51)
        at org.gradle.api.internal.tasks.execution.SkipUpToDateTaskExecuter.execute(SkipUpToDateTaskExecuter.java:62)
        at org.gradle.api.internal.tasks.execution.ResolveTaskOutputCachingStateExecuter.execute(ResolveTaskOutputCachingStateExecuter.java:54)
        at org.gradle.api.internal.tasks.execution.ValidatingTaskExecuter.execute(ValidatingTaskExecuter.java:60)
        at org.gradle.api.internal.tasks.execution.SkipEmptySourceFilesTaskExecuter.execute(SkipEmptySourceFilesTaskExecuter.java:97)
        at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:87)
        at org.gradle.api.internal.tasks.execution.ResolveTaskArtifactStateTaskExecuter.execute(ResolveTaskArtifactStateTaskExecuter.java:52)
        at org.gradle.api.internal.tasks.execution.SkipTaskWithNoActionsExecuter.execute(SkipTaskWithNoActionsExecuter.java:52)
[...]

PR #217, opened 2018-01-26
Build now uses shadow to relocate dependencies.

bigwaff:
I incorporated the shadow plugin and relocated axion's dependencies. Build succeeds and unit/integration tests pass (although I am not sure whether a version of axion plugin with relocated dependencies is used for integration tests). I verified a version of axion with relocated dependencies (using grgit 1.7.2) allows my custom project which declares a dependency on grgit 2.1.0 to use axion w/out issues, i.e. the previous error which prompted this PR does not happen.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,12 +3,12 @@ plugins {
     id 'java-gradle-plugin'
     id 'maven'
     id 'jacoco'
+    	id 'com.github.johnrengelman.plugin-shadow' version '2.0.2'
     id 'pl.allegro.tech.build.axion-release' version '1.8.1'
     id 'com.github.kt3k.coveralls' version '2.3.1'
     id 'com.gradle.plugin-publish' version '0.9.8'
     id 'com.bmuschko.nexus' version '2.3.1' apply false
     id 'io.codearte.nexus-staging' version '0.5.3' apply false
-
     id 'me.champeau.gradle.jmh' version '0.4.3'
     id 'com.bmuschko.docker-remote-api' version '3.2.1'
 }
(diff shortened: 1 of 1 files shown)


adamdubiel:
I see the point, but i'm not sure i want to shadow all dependencies. I would rather make sure things work with newest GrGit.

bigwaff:
> I would rather make sure things work with newest GrGit.

There's a PR #215 for that ;-) However, newest GrGit requires Java 8. Maybe that is undesirable. If so, shadow allows axion-release to remain on Java 7 supported release of GrGit without forcing downstream clients to use latest GrGit and Java 8. Additionally, the classpath pollution not only effects downstream client users but other plugins and their dependencies as well. Relocating axion-release dependencies insulates axion-release, other plugins, and downstream client user build file logic.

Maybe a release of current axion-release plugin using shadow so user who wants to remain on Java 7 can do so w/out classpath pollution? And then follow on release of axion-release plugin with updated dependencies which require Java 8?

bigwaff:
@adamdubiel Have you had a chance to consider my last comment? I have time this weekend to incorporate any feedback you have.

adamdubiel:
@bigwaff i am done with a thing that i wanted to do a long time ago - remove GrGit dependency altogether from production code (but leave it in test code for convenience, at least for some time). This should fix any issues with shadowed dependencies, as the only ones left are `JGit` and `JSch`.

adamdubiel:
PR is now open: #221

adamdubiel:
As to Java 7 vs Java 8: Gradle 4.2 [deprecated the support for Java 7](https://docs.gradle.org/4.2/release-notes.html#java-7-support-is-deprecated).

adamdubiel:
1.9.0 without GrGit is out now - i know it does not solve the problem completly (JSch and JGit are on shared classpath), but does it qualify to close this PR?

adamdubiel:
> 1.9.0 without GrGit is out now - i know it does not solve the problem completly (JSch and JGit are on > shared classpath), but does it qualify to close this PR?

@bigwaff how do you think? ^

bigwaff:
I think this PR is no longer necessary ;-)

PR #221, opened 2018-02-11
Removes GrGit from production dependencies

adamdubiel:


bigwaff:
Can't complain about cutting out the middle man ;-)

PR #227, opened 2018-03-23
Update dependencies

friederbluemle:
A couple of dependency updates:

- Gradle wrapper `4.6`
- `pl.allegro.tech.build.axion-release` `1.9.0`

<sub><sup>Once approved, please use a *normal* GitHub merge (i.e. **NO rebase/squash** merge) to integrate the commit(s) from the branch of this pull request. The changes are broken up into meaningful, [atomic commits](https://www.freshconsulting.com/atomic-commits/), and my branch should already be up-to-date with the latest target branch. If it isn't, or if you'd like me to combine something, please let me know, and I will make the necessary updates as soon as possible. Thank you!</sup></sub>


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
-#Tue Oct 03 19:54:39 CEST 2017
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.6-all.zip
```
adamdubiel: please use -bin :) -all is a waste of bandwith

build.gradle
```diff
@@ -192,10 +192,6 @@ jacocoTestReport {
     }
 }
 
-task wrapper(type: Wrapper) {
-    gradleVersion = '4.2'
```
adamdubiel: why did you remove this section? it was convenient for me to see current version and be able to update Gradle this way

mikezx6r: Wrapper is built-in to Gradle now. If you run gradle wrapper on a new/existing project, it will set the wrapper version to the version of gradle you have installed.

There's an option you can pass to set the version explicitly. `gradle wrapper --gradle-version 4.6`

So these are redundant lines.

The easiest way to see/update the version is to look in the gradle-wrapper.properties file. And that's all the task is doing anyway...

gradle/wrapper/gradle-wrapper.properties
friederbluemle: Sure thing, I changed it to `-bin` - but keep in mind that you're losing out on source support in your IDE when working on Gradle scripts (it can be very helpful). Also - it is a *one-time,  machine-wide* download, normally cached under $HOME/.gradle, so the additional bandwidth requirement only applies *once*. Furthermore, `-all` is the default for new Android Studio projects. Maybe something to consider for the future :)

build.gradle
friederbluemle: Yes, the `wrapper` task is built-in now. Removing these redundant lines will avoid having to modify `build.gradle` during a Gradle wrapper upgrade.

Three easy ways to check the version of the used Gradle wrapper:

1. Open `gradle-wrapper.properties` in an editor, check `distributionUrl`
1. `cat gradle/wrapper/gradle-wrapper.properties | grep Url`
1. `./gradlew --version`

adamdubiel: I like my old ways, but okay, i feel convinced.

adamdubiel:
Thanks a lot for this contribution! :)

PR #237, opened 2018-05-16
Fix two different version commits on the same tag issue

### LINKED ISSUE #236: Two tags on the same commit behavior clarification
When we have one commit with two tags -- say `1.0.0` and `1.1.0-alpha`, the computed version is always the 1.1.0-SNAPSHOT one. We were thinking of having a flow where we release a version, and immediately mark the new one for development. This doesn't seem possible with current behavior, since after marking new release we're not able to infer current stable release.

Our reasoning is that stable version should have higher priority here, since there are no commits made for the next version yet. However I've seen some issues that already requested some specific behavior in this case, so I'd like a clarification on what's the suggested approach and whether there's some option that we can use to achieve what we want.

I'm willing to submit a pull request if it makes sense and isn't a ground-breaking change.

lwasylkowski:
Fixes #236

### DIFF
--- a/src/integration/groovy/pl/allegro/tech/build/axion/release/HighestVersionIntegrationTest.groovy
+++ b/src/integration/groovy/pl/allegro/tech/build/axion/release/HighestVersionIntegrationTest.groovy
@@ -19,7 +19,7 @@ class HighestVersionIntegrationTest extends BaseIntegrationTest {
         result.task(":currentVersion").outcome == TaskOutcome.SUCCESS
     }
 
-    def "should return tag with highest version when there are normal and alpha releases on single commit"() {
+    def "should return tag with stable version when there are normal and alpha releases on single commit"() {
         given:
         buildFile('')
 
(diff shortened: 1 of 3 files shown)


lwasylkowski:
It seems like `HighestVersionIntegrationTest#"should return tag with highest version when there are normal and alpha releases on single commit"` was failing after this fix. It seems like its specification is wrong, though (according to #236) -- from what you say @adamdubiel, the preferred version is stable one, not highest one. I went ahead and modified the test, but if there's anything wrong with this let me know

adamdubiel:
Yes, you are right. IMO the preferred order is:

1) stable version
1) if multiple stable versions, choose higher
1) alpha/nextVersion

Stable version is the most important one, since we probably want to be able to (re-)produce artifacts based on it, even after releasing newer versions.

lwasylkowski:
@adamdubiel I went ahead and added two tests that verify that behaviour

adamdubiel:
I just released 1.9.1.

PR #242, opened 2018-06-11
added regex to filter out tags which overlap in repo

Darkest:
This pull request fixes the following restriction: tag prefixes for each module do not overlap, i.e. !tagA.startsWith(tagB) for each permutation of all tag prefixes

Now tags are filtered by stronger regex, filtering out all tags that do not match the pattern <tagPrefix><tagSeparator>-<semver>


adamdubiel:
This looks really nice, but i'm worried about one thing: user can override tag serialization (using `scmVersion.tag.serialize` property) and chosen tag serialization might not match stronger regexp.. i'm not yet sure if this is a blocker, but to be on a safe side you might add a config flag to turn "strong regexp" on/off. It could be on by default in 1.10.0 . What do you think?

Darkest:
I got your idea. Will look into it as i have time.

PR #268, opened 2018-10-19
Use next release version when force snapshot is used

### LINKED ISSUE #264: CurrentVersion issue: nextVersionTag on same commit as release version and forceSnapshot yield release version
### Issue
When `-Prelease.forceSnapshot` is specified and last tag is `nextVersionTag` on same commit as a release tag, than `currentVersion` is the already released version.

### Test Case
    def "should prefer snapshot of nextVersion when both on current commit and forceSnapshot is enabled"() {
        given:
        repository.tag('release-1.0.0')
        repository.tag('release-1.1.0-alpha')
        VersionProperties versionRules = versionProperties().forceSnapshot().build()
        
        when:
        VersionContext version = resolver.resolveVersion(versionRules, tagRules, nextVersionRules)
        
        then:
        version.previousVersion.toString() == '1.0.0'
        //  also the nextVersion should be taken account when forceSnapshot is specified
        //  BUT IS NOT -- IS 1.0.0-SNAPSHOT !!!
        version.version.toString() == '1.1.0'
        version.snapshot
    }

Theer108:
fixed #264

### DIFF
--- a/src/main/groovy/pl/allegro/tech/build/axion/release/domain/VersionResolver.groovy
+++ b/src/main/groovy/pl/allegro/tech/build/axion/release/domain/VersionResolver.groovy
@@ -37,9 +37,9 @@ class VersionResolver {
 
         Map versions
         if (versionFactory.versionProperties.useHighestVersion) {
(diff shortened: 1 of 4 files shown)


adamdubiel:
Thanks!

PR #277, opened 2018-11-30
fix markNextVersion task documentation

### LINKED ISSUE #276: Documentation of markNextVersion task
docs/cnofiguration/next_version.md:33 uses a task `nextVersion` that does not exist, using a property name that is not used `incrementer=incrementMinor`

rbrindl:
Change the second to last command of the sample session to correct task and property name.
fixes #276


adamdubiel:
Thanks!

Issue #286, opened 2019-03-28
Transition from version to rc version


krzysiekbielicki:
incrementPrerelease should work like snapshot versions - increment patch and add -rc1 in case of no previous prerelease.

```
> ./gradlew currentVersion
1.0.1
> ./gradlew release  -Prelease.versionIncrementer=incrementPrerelease
> ./gradlew currentVersion
1.0.2-rc1
> ./gradlew release  -Prelease.versionIncrementer=incrementPrerelease
> ./gradlew currentVersion
1.0.2-rc2
```

adamdubiel:
I left a comment in PR. I think this would introduce breaking change - i would rahter see new incrementer instead.

grzegorzwitkowski:
hi guys, it doesn't work for me in 1.10.0
having a project with single release tag 0.1.1 and calling
```
gradle-rc1-test (master|✔) $ ./gradlew release -Prelease.dryRun

> Task :verifyRelease
DRY-RUN: uncommitted changes: false
Looking for uncommitted changes..
DRY-RUN: ahead of remote: false
Checking if branch is ahead of remote..
Checking for snapshot versions..

> Task :release
Creating tag: 0.1.1
DRY-RUN: creating tag with name: 0.1.1
Pushing all to remote: origin
DRY-RUN: pushing to remote: origin
```

still produces 0.1.1 but as I understand it should be 0.1.1-rc1 right?

my config

```
plugins {
    id 'pl.allegro.tech.build.axion-release' version '1.10.0'
}

scmVersion {
    tag {
        prefix = ''
    }
    versionIncrementer 'incrementPrerelease', [initialPreReleaseIfNotOnPrerelease: 'rc1']
}

project.version = scmVersion.version
```

adamdubiel:
Try `1.10.1`, it wasn't release in`1.10.0` :) I just published it.

grzegorzwitkowski:
Big thanks, now it works. That's great, I was already afraid I would have to switch to something more reliable from PHP world [;)](https://allegro.tech/2019/04/from-php-to-java-and-back-again.html)

adamdubiel:
Give us some more time to build the essential tooling ;)

PR #287, opened 2019-03-28
[issue #286] Create rc1 in case of no rc version in incrementPrerelase

krzysiekbielicki:
Create rc1 in case of no rc version in incrementPrerelase strategy

### DIFF
--- a/src/main/groovy/pl/allegro/tech/build/axion/release/domain/PredefinedVersionIncrementer.groovy
+++ b/src/main/groovy/pl/allegro/tech/build/axion/release/domain/PredefinedVersionIncrementer.groovy
@@ -43,7 +43,7 @@ enum PredefinedVersionIncrementer {
                     .build()
             }
         }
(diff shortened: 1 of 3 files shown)


src/main/groovy/pl/allegro/tech/build/axion/release/domain/PredefinedVersionIncrementer.groovy
```diff
@@ -43,7 +43,7 @@ enum PredefinedVersionIncrementer {
                     .build()
             }
         }
-        return context.currentVersion.incrementPatchVersion()
+        return context.currentVersion.incrementPatchVersion('rc1')
```
adamdubiel: i wnoder.. this is a breaking change in increment_prerelease behaviour. one might expect, that you can have this always on and it behaves like a normal incrementer unless `rcX` is found. I would rather have a separate incrementer for adding rc1

PR #288, opened 2019-03-28
Java11

krzysiekbielicki:


.travis.yml
```diff
@@ -7,7 +7,7 @@ script: ./gradlew check
 after_success: ./gradlew jacocoTestReport coveralls
 
 jdk:
-  - oraclejdk8
+  - openjdk11
```
adamdubiel: I would rather stay with `openjdk8` for time being - this might be used a lot on Java 8 and I don't want to start using some new APIs by chance

PR #292, opened 2019-05-09
Monorepo support

### LINKED ISSUE #291: Support multiple Gradle builds in one repository (monorepo)
Hi,

in my use case, I would like to use axion release plugin for several independent gradle builds in the same repository. On Jenkins I am using a composite build, that kicks off all the separate builds in the same pipeline.
Currently, I configured different prefixes to have distinct version tags for each project.

The problem is, that a new commit with changes in one of the projects also results in all other projects being released with an incremented version.

I am already working on a pull request for 'monorepo support' in axion release plugin, which would basically replace the condition `isHead` in `VersionResolver`/`VersionSorter` to find out, if the project is on a release tag. Instead, I would check, if there is a release tag on the latest commit, which contained changes in the rootDir of the current Gradle project.

@adamdubiel Can you imagine to include this feature, if I provide a pull request for it? As I am not very familiar with the codebase, I would appreciate a thorough code review.

sradi:
I would like to discuss my quick and dirty solution, that I implemented, before opening the issue for monorepo support.

Fixes #291

### DIFF
--- a/src/main/groovy/pl/allegro/tech/build/axion/release/domain/VersionResolver.groovy
+++ b/src/main/groovy/pl/allegro/tech/build/axion/release/domain/VersionResolver.groovy
@@ -7,7 +7,6 @@ import pl.allegro.tech.build.axion.release.domain.properties.VersionProperties
 import pl.allegro.tech.build.axion.release.domain.scm.ScmPosition
 import pl.allegro.tech.build.axion.release.domain.scm.ScmRepository
 import pl.allegro.tech.build.axion.release.domain.scm.TaggedCommits
-import pl.allegro.tech.build.axion.release.domain.scm.TagsOnCommit
 
 import java.util.regex.Pattern
 
@@ -25,7 +24,7 @@ class VersionResolver {
     private final ScmRepository repository
 
     private final VersionSorter sorter
-    // TODO Convert to Optional<String> ?!
+    private ScmPosition latestChangePosition
     private String projectRootRelativePath
 
     VersionResolver(ScmRepository repository) {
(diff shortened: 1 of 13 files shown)


src/main/groovy/pl/allegro/tech/build/axion/release/domain/VersionSorter.groovy
```diff
@@ -24,20 +24,21 @@ class VersionSorter {
...
         for (TagsOnCommit tagsEntry : taggedCommits) {
             List<String> tags = tagsEntry.tags
 
-            // next version should be igored when tag is on head
+            // next version should be igored when tag is on revision of the latest change in the relevant folder
             // and there are other, normal tags on it
             // because when on single commit on head - normal ones have precedence
             // however, we should take into account next version
             // in case of forced snapshot
-            boolean ignoreNextVersionOnHead = tagsEntry.isHead &&
```
sradi: I don't really understand the meaning of this line. If this variable wasn't there, VersionSorter wouldn't need to be changed at all.

adamdubiel: This fixes one of the corner cases when multiple tags are on the same tag. Imagine commit with tags:

```
abcf124 -> Work in progress on 2.0.0
abcf123 -> release-1.0.0 release-2.0.0-alpha // alpha is nextVersion tag
```

In following situations expected outcomes are:

```
git checkout abcf124
./gradlew cV
2.0.0-SNAPSHOT # I'm working on next version

git checkout abcf123
./gradlew cV
1.0.0 # I'm on released commit now
```

Thus check for `isHead` is necessary for version sorting. I don't know if this is necessary in your case, because it serves different purpose than establishing if I'm on SNAPSHOT or not.

adamdubiel:
The overall approach looks ok, I'm actually pleasantly surprised how little code had to be changed. I hope maybe with my clarification there is no need for changes in VersionSorter.

src/main/groovy/pl/allegro/tech/build/axion/release/domain/VersionResolver.groovy
```diff
@@ -24,16 +24,27 @@ class VersionResolver {
...
+    VersionResolver(ScmRepository repository, String projectRootRelativePath) {
         this.repository = repository
+        this.projectRootRelativePath = projectRootRelativePath
         this.sorter = new VersionSorter()
     }
 
     VersionContext resolveVersion(VersionProperties versionRules, TagProperties tagProperties, NextVersionProperties nextVersionProperties) {
-        ScmPosition position = repository.currentPosition()
+        if (projectRootRelativePath != null) {
+            latestChangePosition = repository.positionOfLastChangeIn(projectRootRelativePath);
```
sradi: Unfortunately I cannot call methods on repository in the constructor of VersionResolver and set "latestChangePosition" there. 
In the tests, the constructor is called in setup(), before any commits exist in the repo.

src/main/java/pl/allegro/tech/build/axion/release/domain/scm/TaggedCommits.java
```diff
@@ -0,0 +1,42 @@
...
+        TagsOnCommit latestTags = repository.latestTags(tagPattern);
+        return new TaggedCommits(latestTagPosition, Arrays.asList(latestTags));
+    }
+
+    public static TaggedCommits fromAllCommits(ScmRepository repository, Pattern tagPattern, ScmPosition latestTagPosition) {
+        List<TagsOnCommit> taggedCommits = repository.taggedCommits(tagPattern);
+        return new TaggedCommits(latestTagPosition, taggedCommits);
+    }
+
+    public static TaggedCommits fromLatestCommitBeforeNextVersion(ScmRepository repository, Pattern releaseTagPattern, Pattern nextVersionTagPattern, ScmPosition latestTagPosition) {
```
sradi: Any suggestions for better method names are welcome :)

PR #297, opened 2019-07-18
Support fetching authorization from netrc for HTTP protocol

### LINKED ISSUE #296: Support getting credentials from .netrc (or _netrc for Windows)
For SSH, if no authorization option is specified, the plugin will try to fetch authorization options from SSH agent.

Could the plugin do the same for HTTP and try to fetch the authorization if no options are specified, i.e. from netrc file.

hugares:
When no authorization option is configured, fetch authorization from
netrc file. Use JGit NetRCCredentialsProvider which will load
authorization from <user-home>/.netrc on Linux and from
<user-home>\_netrc on Windows.

Resolves #296

### DIFF
--- a/docs/configuration/authorization.md
+++ b/docs/configuration/authorization.md
@@ -3,7 +3,8 @@
 ## SSH Agent
 
 If no other options are used and repository requires authorization, `axion-release`
-will try to fetch authorization options from SSH agent (`ssh-agent` on Linux and `pageant` on Windows).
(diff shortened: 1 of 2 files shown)


hugares:
I tested manually that it works thus making this PR ready for review.

Let me know if you have an idea how I could add test for the new functionality. I did not find where the default authorization for ssh get tested.

adamdubiel:
Great, thanks for this contribution :) No tests for SSH are provided, as it proved to be really really hard to test and since the code is not changing much, after giving it a try I decided to leave it as is.

PR #307, opened 2019-12-04
Remove prefix restriction.

john-tipper:
Prefixes were hitherto restricted from not colliding, i.e. prefixes
`pre1` and `pre2` were forbidden.
Bumped jacoco version to allow for compilation using JDK8.

### DIFF
--- a/docs/configuration/basic_usage.md
+++ b/docs/configuration/basic_usage.md
@@ -83,8 +83,6 @@ Sometimes it might be desirable to release each module (or just some
 modules) of multi-module project separately. If so, please make sure
 that:
 
--   tag prefixes for each module do not overlap, i.e.
-    `!tagA.startsWith(tagB)` for each permutation of all tag prefixes
 -   keep in mind, that `scmVersion` must be initialized before
     `scmVersion.version` is accessed
 -   apply plugin on each module that should be released on it\'s own
--- a/docs/configuration/basic_usage.md
+++ b/docs/configuration/basic_usage.md
@@ -85,7 +85,37 @@ that:
 
 -   keep in mind, that `scmVersion` must be initialized before
     `scmVersion.version` is accessed
--   apply plugin on each module that should be released on it\'s own
+-   apply plugin on each module that should be released on its own
+
+The plugin ignores tags based on the prefix and version separator
+(i.e. `<prefix><separator>`).  If there is a prefix configured then
+tags that do not start with the configured `<prefix><separator>` are 
+all ignored. If there is no prefix configured then only tags matching
+the version are used for calculating the version (i.e. the version 
+separator is also ignored).
+
+This allows for using the name of the module with an appropriate version
+separator as a namespace in a multi-module project, as shown in the table
+below:
+
+| Module name | Version separator | Tags will appear as |
+|-------------|-------------------|---------------------|
+| `module`    | `-`               | `module-<maj>.<min>.<patch>` |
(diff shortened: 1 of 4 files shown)


agusmba:
Just my 2c, sorry if I'm wrong but the prefix restriction AFAIR was a bit different:

> tag prefixes for each module do not overlap, i.e.	
    `!tagA.startsWith(tagB)` for each permutation of all tag prefixes

That means prefix `projectA` and `projectAmodB` were not allowed.

It seems you are testing a different thing (projectA vs projectB), which should work anyway.

john-tipper:
Ah, I see what you mean, good point.  I have changed the test to reflect that, as I want to allow the use of a shared prefix, e.g. `module` and `moduleV2`.

Note that the build failure appears to be nothing I have done, it's a failure of Travis to install JDK8.

src/test/groovy/pl/allegro/tech/build/axion/release/domain/VersionResolverTest.groovy
```diff
@@ -20,6 +20,24 @@ class VersionResolverTest extends RepositoryBasedTest {
 
     VersionProperties defaultVersionRules = versionProperties().build()
 
+    @Shared
```
adamdubiel: i would rather have those created in test, close to the usage context instead of class level. also they only differ in `prefix`, so you can add prefix as parameter of newly created test and all will be much more readable :)

docs/configuration/basic_usage.md
```diff
@@ -83,8 +83,6 @@ Sometimes it might be desirable to release each module (or just some
 modules) of multi-module project separately. If so, please make sure
 that:
 
--   tag prefixes for each module do not overlap, i.e.
```
adamdubiel: while your change makes axion more flexible, it does not solve all problems. This warning was created because of common scenario with projects like:

```
my-service
my-service-client
```

Because version separator and character used in names of projects is the same, it will still fail. I think you should change this warning to illustrate it instead of removing it altogether.

src/test/groovy/pl/allegro/tech/build/axion/release/domain/VersionResolverTest.groovy
john-tipper: You're quite right, I have revised the test.

docs/configuration/basic_usage.md
john-tipper: I have expanded the docs by adding in an explicit warning for this issue.

src/test/groovy/pl/allegro/tech/build/axion/release/domain/VersionResolverTest.groovy
```diff
@@ -4,7 +4,7 @@ import pl.allegro.tech.build.axion.release.RepositoryBasedTest
 import pl.allegro.tech.build.axion.release.domain.properties.NextVersionProperties
 import pl.allegro.tech.build.axion.release.domain.properties.TagProperties
 import pl.allegro.tech.build.axion.release.domain.properties.VersionProperties
-import spock.lang.Ignore
+import spock.lang.Shared
```
adamdubiel: nitpick - this is now unused :) other than that +1

john-tipper: Sorry, fixed!

adamdubiel:
Done, thanks! :)

PR #310, opened 2019-12-05
Monorepo support

john-tipper:
This is a reworked version of the initial work done in #291 by @sradi and his PR #292.  To be properly useful, this should make use of the changes in PR #307.

Note that we want to enable the user to exclude changes in submodules from being considered when calculating the version of the parent project.  When determining whether a commit takes place in a submodule, we use JGit's `addPath` command to only look at commits in a particular subdirectory.  However, JGit does not (yet) support excluding directories from its [LogCommand](https://download.eclipse.org/jgit/site/5.5.1.201910021850-r/apidocs/org/eclipse/jgit/api/LogCommand.html).  This is discussed in a StackOverflow question I asked [here](https://stackoverflow.com/questions/59143934/java-jgit-how-to-get-git-commits-not-affecting-certain-directories).

I have created a PR to JGit (https://git.eclipse.org/r/#/c/153673/) where I add a `excludePath()` method to `LogCommand`, which will allow us to exclude directories from consideration when determining whether to increment a project's version.  When this PR is merged and there is a new version released, we should be able to bump the JGit version and the code [in this section](https://github.com/john-tipper/axion-release-plugin/blob/monorepo-support/src/main/groovy/pl/allegro/tech/build/axion/release/infrastructure/git/GitRepository.groovy#L187-L195) should work (subject to the ignored tests in [MultiModuleProjectIntegrationTest](https://github.com/john-tipper/axion-release-plugin/blob/monorepo-support/src/integration/groovy/pl/allegro/tech/build/axion/release/MultiModuleProjectIntegrationTest.groovy)).  Adding the new version of JGit will require the bump to `grgit-core` to support Jgit v5 that you see in `build.gradle`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -47,7 +47,7 @@ repositories {
 }
 
 project.ext.versions = [
-        jgit: '5.5.1.201910021850-r',
+        jgit: '5.6.0.201912101111-r',
         jsch: '0.1.54',
         jschAgent: '0.0.9'
 ]
@@ -99,7 +99,7 @@ dependencies {
 
     compile group: 'com.github.zafarkhaja', name: 'java-semver', version: '0.9.0'
 
-    testCompile (group: 'org.ajoberstar.grgit', name: 'grgit-core', version: '3.1.1') {
+    testCompile (group: 'org.ajoberstar.grgit', name: 'grgit-core', version: '4.0.1') {
         exclude group: 'org.eclipse.jgit', module: 'org.eclipse.jgit.ui'
         exclude group: 'org.eclipse.jgit', module: 'org.eclipse.jgit'
     }
--- a/build.gradle
+++ b/build.gradle
@@ -16,6 +16,13 @@ if (project.hasProperty('mavenCentral')) {
     apply from: 'gradle/mavenCentral.gradle'
 }
 
+sourceSets {
+    main {
+        java { srcDirs = [] }    // no source dirs for the java compiler
+        groovy { srcDirs = ["src/main/java", "src/main/groovy"] }  // compile   everything in src/ with groovy
+    }
+}
+
 scmVersion {
     tag {
         prefix = 'axion-release'
@@ -36,10 +43,11 @@ sourceCompatibility = '1.7'
 
 repositories {
     mavenCentral()
+    jcenter()
 }
 
 project.ext.versions = [
-        jgit: '4.11.0.201803080745-r',
+        jgit: '5.5.1.201910021850-r',
         jsch: '0.1.54',
         jschAgent: '0.0.9'
 ]
@@ -91,7 +99,7 @@ dependencies {
 
     compile group: 'com.github.zafarkhaja', name: 'java-semver', version: '0.9.0'
 
-    testCompile (group: 'org.ajoberstar', name: 'grgit', version: '1.7.2') {
+    testCompile (group: 'org.ajoberstar.grgit', name: 'grgit-core', version: '3.1.1') {
         exclude group: 'org.eclipse.jgit', module: 'org.eclipse.jgit.ui'
         exclude group: 'org.eclipse.jgit', module: 'org.eclipse.jgit'
     }
@@ -190,7 +198,7 @@ gradlePlugin {
 }
 
 jacoco {
-    toolVersion = '0.8.1'
+    toolVersion = '0.8.2'
 }
 
 jacocoTestReport {
--- a/build.gradle
+++ b/build.gradle
@@ -16,13 +16,6 @@ if (project.hasProperty('mavenCentral')) {
     apply from: 'gradle/mavenCentral.gradle'
 }
 
-sourceSets {
-    main {
-        java { srcDirs = [] }    // no source dirs for the java compiler
-        groovy { srcDirs = ["src/main/java", "src/main/groovy"] }  // compile   everything in src/ with groovy
-    }
-}
-
 scmVersion {
     tag {
         prefix = 'axion-release'
@@ -53,6 +46,11 @@ project.ext.versions = [
 ]
 
 sourceSets {
+    main {
+        java { srcDirs = [] }    // no source dirs for the java compiler
+        groovy { srcDirs = ["src/main/java", "src/main/groovy"] }  // compile   everything in src/ with groovy
+    }
+
     integration {
         java.srcDir project.file('src/integration/java')
         groovy.srcDir project.file('src/integration/groovy')
--- a/docs/configuration/basic_usage.md
+++ b/docs/configuration/basic_usage.md
@@ -77,7 +77,7 @@ allprojects {
 }
 ```
 
-### Multi-module with multiple versions
+### Multi-module project with multiple versions
 
 Sometimes it might be desirable to release each module (or just some
 modules) of multi-module project separately. If so, please make sure
@@ -117,14 +117,16 @@ my-service
 my-service-client
 ```
 
-Use the `foldersToExclude` configuration parameter to identify submodules
+Use the `projectDirs` configuration parameter within a `monorepos` block to identify submodules
 that should be excluded from consideration when calculating whether to increment
 the version of the parent project.  Typically, you would do this in the top level
 project, assuming that submodules are named after the directory they appear in:
 
 ```
 scmVersion {
-    foldersToExclude = project.subprojects.collect({p -> p.name})
+    monorepos {
+        projectDirs = project.subprojects.collect({p -> p.name})
+    }
 }
 ```
 
--- a/docs/configuration/basic_usage.md
+++ b/docs/configuration/basic_usage.md
@@ -89,6 +89,26 @@ that:
     `scmVersion.version` is accessed
 -   apply plugin on each module that should be released on it\'s own
 
+Use the `foldersToExclude` configuration parameter to identify submodules
+that should be excluded from consideration when calculating whether to increment
+the version of the parent project.  Typically, you would do this in the top level
+project, assuming that submodules are named after the directory they appear in:
+
+```
+scmVersion {
+    foldersToExclude = project.subprojects.collect({p -> p.name})
+}
+```
+
+Version calculation rules:
+1. Changes to files within a submodule increment that submodule's version only.
+2. Changes to a submodule do not cause a change to the parent project's version if
+the parent is set to ignore that submodule, via `foldersToExclude`.
+3. Changes to files in the parent project but which are not in a submodule identified via 
+`foldersToExclude` will cause the parent project's version to increment but not the 
+versions of any submodules.  If this is desired then consider wiring the `createRelease` or
+`release` tasks of the submodules to be dependencies of the tasks of the same name in the parent. 
+
 ## Releasing
 
 ```
--- a/src/main/groovy/pl/allegro/tech/build/axion/release/domain/VersionConfig.groovy
+++ b/src/main/groovy/pl/allegro/tech/build/axion/release/domain/VersionConfig.groovy
@@ -49,7 +49,7 @@ class VersionConfig {
 
     HooksConfig hooks = new HooksConfig()
 
-    List<String> foldersToExclude = []
+    MonorepoConfig monorepoConfig = new MonorepoConfig()
 
     private Context context
 
@@ -166,5 +166,9 @@ class VersionConfig {
         if (context == null) {
             this.context = GradleAwareContext.create(project, this)
         }
+    }
+
+    void monorepos(Closure c) {
+        project.configure(monorepoConfig, c)
     }
 }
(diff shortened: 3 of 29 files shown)


john-tipper:
Update: my PR (https://git.eclipse.org/r/#/c/153673/) was merged into JGit last night.  It should feature in the 5.6 release, not sure on a timetable for that yet.

john-tipper:
Further update on PR https://git.eclipse.org/r/#/c/153673/ below from one of the project's maintainers:

> final release build for 5.6 is next Tuesday or Wednesday. It will be officially published with new Eclipse simultaneous release 2019-12 on Dec 18

john-tipper:
5.6 artifact will be available on 18 Dec, so I will attempt to progress the PR at that point.
https://projects.eclipse.org/projects/technology.jgit/releases/5.6.0/plan

john-tipper:
The new version of JGit has been incorporated, adding support for excluding directories when calculating changes, plus a new version of GrGit, which was needed to support the new JGit version.

I have removed the "Draft" status of the PR, as I think this is ready to be looked at with a view to merging; happy to accept any/all feedback.

adamdubiel:
Thanks, I will take a look at this :)

src/main/groovy/pl/allegro/tech/build/axion/release/domain/VersionConfig.groovy
```diff
@@ -49,6 +49,8 @@ class VersionConfig {
 
     HooksConfig hooks = new HooksConfig()
 
+    List<String> foldersToExclude = []
```
adamdubiel: i wonder if we should use "directory" instead of folder (`directory` is already used in `RepositoryConfig`), so maybe `dirsToExclude`?

src/main/groovy/pl/allegro/tech/build/axion/release/infrastructure/di/Context.groovy
```diff
@@ -31,7 +31,20 @@ class Context {
 
         instances[ScmRepository] = scmRepository
         instances[VersionService] = new VersionService(new VersionResolver(scmRepository))
+    }
+
+    public Context(Properties rules, ScmRepository scmRepository, ScmProperties scmProperties, File projectRoot, LocalOnlyResolver localOnlyResolver, VersionConfig versionConfig) {
```
adamdubiel: the idea behind having `Context` and `GradleAwareContext` classes separate was to isolate any Gradle-dependent code from plugin domain. The isolation is not perfect, but i would like to try and keep it this way. `VersionConfig` object in my view is part of Gradle config (it is mutable etc), and so it should not be imported by `Context` class. Can you change it? I know it will make constructor uglier, so maybe some `Properties` object is needed.

adamdubiel: also is old constructor needed? why not keep only one, extended constructor?

src/main/groovy/pl/allegro/tech/build/axion/release/domain/VersionResolver.groovy
```diff
@@ -24,16 +24,33 @@ class VersionResolver {
     private final ScmRepository repository
 
     private final VersionSorter sorter
+    private ScmPosition latestChangePosition
+    private String projectRootRelativePath
+    private List<String> foldersToExcludeWhenResolvingLatestRelevantRootProjectCommit
 
     VersionResolver(ScmRepository repository) {
+        this(repository, null, Collections.emptyList())
```
adamdubiel: if we need only one constructor - go ahead and remove this one :)

src/main/groovy/pl/allegro/tech/build/axion/release/domain/VersionResolver.groovy
```diff
@@ -24,16 +24,33 @@ class VersionResolver {
...
+    VersionResolver(ScmRepository repository, String projectRootRelativePath, List<String> excludeFolders) {
         this.repository = repository
+        this.projectRootRelativePath = projectRootRelativePath
         this.sorter = new VersionSorter()
+        this.foldersToExcludeWhenResolvingLatestRelevantRootProjectCommit = excludeFolders
     }
 
     VersionContext resolveVersion(VersionProperties versionRules, TagProperties tagProperties, NextVersionProperties nextVersionProperties) {
-        ScmPosition position = repository.currentPosition()
+        if (projectRootRelativePath != null) {
```
adamdubiel: from what i see `projectRootRelativePath` is never `null` on production (might be null in tests) - why keep this switch? it makes prod code more complicated and i don't want tests to complicate prod code

build.gradle
```diff
@@ -16,6 +16,13 @@ if (project.hasProperty('mavenCentral')) {
     apply from: 'gradle/mavenCentral.gradle'
 }
 
+sourceSets {
```
adamdubiel: is this change necessary?

docs/configuration/basic_usage.md
```diff
@@ -117,6 +117,26 @@ my-service
 my-service-client
 ```
 
+Use the `foldersToExclude` configuration parameter to identify submodules
```
adamdubiel: I think this prominent feature should be also more visible in configuration. In general the long-term maintenance of this would be easier if there was:

* separate configuration object (features might pile up in the future)
* feature switch to allow using old flow without `null` or `isEmpty` checks

I suggest creating `MonorepoConfig` (transformed into `MonorepoProperties` to keep Gradle/domain isolation):

```
class MonorepoConfig {
    boolean enabled = false;
    List<String> projectDirs = [];
}

class VersionConfig {
    MonorepoConfig monorepo;
}
```

src/main/groovy/pl/allegro/tech/build/axion/release/domain/VersionResolver.groovy
adamdubiel: also: this is the place where old vs new flow can be distinguished using feature switch i proposed :)

src/main/groovy/pl/allegro/tech/build/axion/release/domain/VersionResolver.groovy
```diff
@@ -64,23 +81,22 @@ class VersionResolver {
         boolean forceSnapshot = versionProperties.forceSnapshot
 
         Map currentVersionInfo, previousVersionInfo
-        TagsOnCommit latestTags = repository.latestTags(releaseTagPattern)
-        currentVersionInfo = versionFromTaggedCommits([latestTags], false, nextVersionTagPattern,
+        TaggedCommits latestTaggedCommit = TaggedCommits.fromLatestCommit(repository, releaseTagPattern, latestChangePosition)
+        currentVersionInfo = versionFromTaggedCommits(latestTaggedCommit, false, nextVersionTagPattern,
             versionFactory, forceSnapshot)
+        boolean onCommitWithLatestChange = currentVersionInfo.commit == latestChangePosition.revision
+        println("currentVersion = ${currentVersionInfo.commit}, latestChangePosition=${latestChangePosition.revision}, onCommitWithLatestChange=${onCommitWithLatestChange}, currentVersionInfo.isNextVersion=${currentVersionInfo.isNextVersion}")
```
adamdubiel: should it stay here? :)

john-tipper:
Thanks very much, let me dig into these get back to you with changes asap.

src/main/groovy/pl/allegro/tech/build/axion/release/domain/VersionConfig.groovy
john-tipper: Agreed.

build.gradle
john-tipper: Yes, didn't see the second block when I added this block, so have merged with the `sourceSets` block further down in the file.

src/main/groovy/pl/allegro/tech/build/axion/release/domain/VersionResolver.groovy
john-tipper: Sorry, I blame a tired programmer working late, personally :-)  I've removed it.

src/main/groovy/pl/allegro/tech/build/axion/release/infrastructure/di/Context.groovy
john-tipper: Please would you mind confirming your view on what `Context` should do - it's a class to encapsulate all the Git-specific classes/tools etc, right?  So `Context` should have some sort of properties (`MonoRepoProperties`) that are used to process mono-repos, and this class would be created via a similar static method to the call to `LocalOnlyResolverFactory.create(...)`?

e.g. 
`MonoRepoProperties.createFromConfig(MonoRepoConfig config)
`
where `MonoRepoconfig` is a class contained within `VersionConfig`?

adamdubiel: `Context` is a class to build the whole core environment - meaning version reading & resolution. Core logic should have nothing to do with Gradle.

Each `*Config` file is transformed into `*Properties` using a separate `*Factory`. Factory transforms Gradle domain objects (`*Config`) into core domain (`*Properties`) while handing cmd args and env overwrites. `*Factory` classes belong to Gradle, as they are aware of Gradle.

So, in general, the second answer :)

john-tipper: Amended.

src/main/groovy/pl/allegro/tech/build/axion/release/domain/VersionResolver.groovy
john-tipper: Amended.

src/main/groovy/pl/allegro/tech/build/axion/release/domain/VersionResolver.groovy
john-tipper: Yes the check was to see if the path was `null` as a result of tests, which it no longer can be due to the changes of constructor.  However, all resolution of versions must go through the new method, as the project may be running as a submodule, i.e. `:submodule:currentVersion`, as opposed to `:currentversion`.  The configuration parsed by that submodule is its own config and not that of the parent project, which is where the feature flag to enable monorepo support would be set (i.e. the subprojects would also need to have the feature flag set to true in their own config). I don't think it's correct that submodules should have to also have the flag set. This means that there's no need for a feature flag.

john-tipper:
I've made some more changes, so hopefully it's closer to what you'd like.  The only thing I didn't implement is the feature toggle - this is because even if a project does not have a enable flag set, that project may well be running as a submodule of a parent project which does have the flag set.

src/main/groovy/pl/allegro/tech/build/axion/release/domain/properties/VersionProperties.groovy
```diff
@@ -17,9 +17,11 @@ class VersionProperties {
     final Closure<Version> versionIncrementer
 
     final boolean sanitizeVersion
-    
+
     final boolean useHighestVersion
 
+    final MonorepoProperties monorepoProperties
```
adamdubiel: i think `MonorepoProperties` deserves to be a top-level `Properties` object and doesn't have to be nested in `VersionProperties`, but since this is internal object, it can be changed at any time :)

adamdubiel:
You are right about feature flag, i didn't take this into a count. Everything looks good to me, added one minor thing. If you feel like changing it it would be great. If not - I'm willing to merge it anyways. Lots of good work and great tests in here :) This will be shipped as 1.11.

src/main/groovy/pl/allegro/tech/build/axion/release/domain/properties/VersionProperties.groovy
john-tipper: `VersionService` currently receives in `currentVersion()` and `currentDecoratedVersion()` a `VersionProperties` object.  To me, this seems quite clean, in that `VersionService` only receives properties that are related to itself and its behaviour is not changed by properties outside of its concern (e.g. it doesn't receive a `HooksProperties` object, for instance).  `VersionService` in turn passes these properties into `VersionResolver`.

If we were to move `MonorepoProperties` out of `VersionProperties` and into `Properties` then `VersionService` would need to have passed to it either the full `Properties` object or both  `MonorepoProperties` and `VersionProperties`.  I'm happy to make the change if you'd like, but it seems to me to result in somewhat more unwieldy service classes - what do you think?

adamdubiel: I feel convinced. Let's keep it as it is :)

john-tipper: Thanks :-)

john-tipper:
Thanks very much for the code reviews, it was fun getting the design to work nicely!  When do you expect to push the 1.11.0 release?

adamdubiel:
Probably on Friday or during the weekend :) I want to play around with monorepo on snapshot builds.

john-tipper:
Thanks for merging, I've updated the docs to refer to the new name for the monorepo configuration block in a separate PR #318 (`foldersToExclude` was changed to `projectDirs`).  Sorry, I missed this in this PR.

Issue #311, opened 2019-12-05
Add parameter to pass current branch name in case repo is in detached state


maniekq:
This is feature request. I would like to know if adding this functionality makes sense or I am missing some existing possibilities.
Aim of this functionality would be to have additional parameter like `-Prelease.branchName` which would allow to override/set branch name read from Git (or not set in case repo is in detached state).  

I will try to describe my use case.
I am using Azure DevOps Pipelines to build my application. This CI system check out Git repo in detached head state. Because of that I am always running axion release with parameters `-Prelease.disableChecks` and  `-Prelease.pushTagsOnly` and it works great for simple cases.
However, when I want to add configuration per branch like the one below, it will not work, because branch is not known.
```
scmVersion {
	useHighestVersion = true
	branchVersionCreator = [
		'hotfix/.*': 'simple',
		'feature/.*': 'versionWithBranch'
	]
	versionIncrementer 'incrementMinor'
	branchVersionIncrementer = [
		'hotfix/.*': 'incrementPatch',
	]
}
```

If I could set branch name with `-Prelease.branchName=$(release.branch.name)`, I would be able to pass it from Azure DevOps and use full functionality of axion-release-plugin.

maniekq:
@adamdubiel I would be interested in implementing this, but I would like to know if this makes sense from your perspective.

adamdubiel:
Yes, please do, i know dealing with CI/CD can be a pain so anything that makes it easier is welcome. It would be ideal if you described your Azure Pipelines setup in [docs](https://axion-release-plugin.readthedocs.io/en/latest/configuration/ci_servers/) as well :)

PR #313, opened 2019-12-29
Introducing `overriddenBranchName` flag

maniekq:
To be able to use branch-related features when CI is using detached head state of Git repository
This is done to solve issue #311.


maniekq:
I was looking at various ways of implementing this and decided to go with the simplest - overriding branch name in GitRepository class. I believe this is also the most reliable way, because it should work with all existing features and those added in the future.

adamdubiel:
Thanks for the contribution! LGTM, going to merge and release this one

Issue #314, opened 2019-12-31
Follow the dependencies between the submodules during the release


kboom:
Maintaining separate versions for sub-projects is very useful but then releasing must be done with care - projects with internal dependencies need to be released manually one after the other - dependencies first. It would be great to:

1. Make the release task understand this and release the dependencies (if necessary) before attempting to release a selected module
2. Make the release task release all modules of a gradle project according to a safe order.

Currently typing `-Prelease.useHighestVersion -Prelease.disableChecks -Prelease.pushTagsOnly` on the top level project won't work:

> The project uses snapshot versions - can't release. Snapshots found: ...

Doing that from a scope of a project with dependencies won't work either. Of course this happens only when we use `dependency(project(":submodule"))` notation rather than `dependency("my project:mymodule:x.y.z)` which would be difficult to maintain

adamdubiel:
Looks like task dependencies problem, can't it be achieved using custom task in Gradle with specified dependencies?

john-tipper:
I have addressed this issue in PR #326.

PR #326, opened 2020-01-29
#325 #314 Enable multi-module project lib dependency incrementing.

### LINKED ISSUE #325: Need to be able to force increment a release
With the new support for monorepos, there is an issue when a submodule depends on another module but there are no code changes within the dependent module, in that the dependent module does not have its version incremented.

For example, given a project with modules `mod_a` and `mod_b`, where both `mod_a` and `mod_b` are to be published as artifacts and also where `mod_b` is dependent on `project(":mod_a")`, then a change within `mod_a` will result in the version of `mod_a` being incremented, but the version of `mod_b` will not.

Within the `Releaser` class (https://github.com/allegro/axion-release-plugin/blob/master/src/main/java/pl/allegro/tech/build/axion/release/domain/Releaser.java#L31), there is a check to see if there are changes and if not (i.e. the version is not a snapshot) then the version will not increment.  There does not appear to me to be an easy way to request that `mod_b`'s version be incremented too (as opposed to forcing it to be a specific version via `-Prelease.forceVersion=x.y.z`).

Gradle has a `buildNeeded` and `buildDependents` task (c.g. [docs here](https://docs.gradle.org/current/userguide/multi_project_builds.html#sec:multiproject_build_and_test)), where, for the example above, I could call `:mod_a:buildDependents` and `:mod_a:build` and `:mod_b:build` would be called.

Ideally, I think under CI/CD it would be nice to just call `./gradlew release` and have the version(s) of the modules that changed be incremented (where we count a dependency between the modules as being a material change in the dependent module), even if there were no code changes to the code actually within the dependent modules.

I think this is loosely related to #314.

Whilst this sort of logic does not necessarily need to reside inside this plugin, it would be nice to document within the project how this sort of functionality could be achieved.

@adamdubiel, @kboom and @sradi, what are your thoughts here please?

john-tipper:
Fixes #325 and #314: enables modules dependent on other submodules to have their version incremented, even if there are no code changes in that module.

There were two issues with the plugin when it handled multi-module projects:
- modules that depended on other modules did not have their version incremented if there were no code changes within that module.  There is no mechanism for force incrementing the version of a module.
- assuming that there were a mechanism for force incrementing a module's version given that its dependent modules' versions have changed, the release tag is placed on HEAD, not on the last relevant commit for that module (the commit closest to HEAD that occurred with changes within that module's directory).

This PR fixes these issues in a non-breaking manner.

The PR adds 2 additional top-level tasks, `createRelaseDependents` and `releaseDependents`, which are analogous to `createRelease` and `release` respectively.  For the purposes of this PR description, `releaseDependents` will be used, but `createReleaseDependents` operates in a similar manner.  The purpose of `releaseDependents` is to create a release for a module that has changed and also for all modules that rely on that module, and for this change to then be transitively cascaded down the dependency tree.

The current `release` mechanism is idempotent and this is a very nice feature and must be retained. Calling `releaseDependents` repeatedly should not cause multiple releases to be generated.

The `releaseDependents` task operates in an identical fashion to Gradle's Java plugin, which adds `buildNeeded` and `buildDependents` tasks, as described [here](https://docs.gradle.org/current/userguide/multi_project_builds.html#sec:multiproject_build_and_test).  Within each project, the `release` task is added as a dependency of `releaseDependents`, and that project's `releaseDependents` task is added as a dependency of all downstream projects.  Calling `releaseDependents` will thus cause `release` to be called in all downstream dependent projects.

In order to actually cause the release to be created, it is necessary for the `release` task to take account of whether upstream releases were created.  The plugin adds a task called `configureReleaseTasksTask` that is set as a dependency of `releaseDependents` and to run prior to `release`.  Calling `releaseDependents` will thus cause the following execution order: `configureReleaseTasksTask`, `release`, `releaseDependents`.  This task checks upstream tasks of the same name and determines if they released a new release by looking at an `ext` property on that task.  If any upstream task caused a release then the task will configure the `release` task to create a version.

We don't want a user to be able to mess up the idempotency of the release process through poor config, so there is a property within `VersionProperties` that determines whether a version should be force incremented (it is not parsed from config).  That property is set based on a flag in the `ReleaseTask` that is set by the `configureReleaseTasksTask` that runs before it.

If a project has changes then the normal release process will be followed.  If there are no changes then a release is created if the force increment flag is set. In this case, the new release tag is written to the last commit that relates to that project, which is not necessarily HEAD.  This ensures that the version resolution for these projects is correct as they are set up to ignore changes in directories that are not related to the project in question.

Here is an example of one of the `MultiModuleProjectDependencyIntegrationTest` unit tests showing where release tags are placed using this new process:

```
commit 8b8866eded61b593442403a058d91816893d043b (HEAD -> refs/heads/master, tag: refs/tags/project1-2.0.1)
  Author: John Tipper <john.tipper@example.com>
  Date:   Tue Jan 28 22:57:34 2020 +0000
  
      commit submodule project1
  
  commit 87ff9829a1b123dcec37f97998556a42d9e72b9e (tag: refs/tags/project3-4.0.1, tag: refs/tags/project3-4.0.0)
  Author: John Tipper <john.tipper@example.com>
  Date:   Tue Jan 28 22:57:34 2020 +0000
  
      commit submodule project3
  
  commit 6cd13c3a838d80c87761d6afe27f8f4566f1d372 (tag: refs/tags/project2-3.0.1, tag: refs/tags/project2-3.0.0)
  Author: John Tipper <john.tipper@example.com>
  Date:   Tue Jan 28 22:57:33 2020 +0000
  
      commit submodule project2
  
  commit 540259aea51bea5b95f483ad67d0cc05be44f4da (tag: refs/tags/project1-2.0.0)
  Author: John Tipper <john.tipper@example.com>
  Date:   Tue Jan 28 22:57:33 2020 +0000
  
      commit submodule project1
  
  commit 1d89f8ba50b5a6f0e37947ea2594272f8edb40fe (tag: refs/tags/release-1.0.0)
  Author: John Tipper <john.tipper@example.com>
  Date:   Tue Jan 28 22:57:33 2020 +0000
  
      initial commit of top level project
  
  commit 949dffd4afcafa63ef280670ab26e617aa2ffdbf
  Author: John Tipper <john.tipper@example.com>
  Date:   Tue Jan 28 22:57:33 2020 +0000
  
      initial commit
```

### DIFF
--- a/docs/configuration/basic_usage.md
+++ b/docs/configuration/basic_usage.md
@@ -83,7 +83,7 @@ Sometimes it might be desirable to release each module (or just some
 modules) of multi-module project separately. If so, please make sure
 that:
 
--   keep in mind, that `scmVersion` must be initialized before
+-   you keep in mind that `scmVersion` must be initialized before
     `scmVersion.version` is accessed
 -   apply plugin on each module that should be released on its own
 
@@ -105,6 +105,16 @@ below:
 
 For example, within `module`, tags that do not start `module-` will be
 ignored. 
+
+Thus we can configure each subproject within such a multimodule project like this:
+
+```
+scmVersion {
+    tag {
+        prefix = "${project.name}"
+    }
+}
+```
 
 **IMPORTANT:**  
 
@@ -150,3 +160,5 @@ versions of any submodules.  If this is desired then consider wiring the `create
 # ./gradlew cV
 0.1.0
 ```
+
+For a multimodule project which has project lib dependencies (e.g. `moduleB` `dependsOn` `project(':moduleA')`), a change within `moduleA`'s code will result in the version of `moduleA` being incremented normally.  However, `moduleB`'s code may not have changed and thus it will not have its version incremented, which is usually not the desired behaviour.  This issue is rectified by the 2 tasks, `createReleaseDependents` and `releaseDependents`.  These tasks are analogous to `createRelease` and `release`, respectively, and traverse the inter-project dependency tree and cause the creation of releases for all projects that have a declared dependency on a project(s) whose code has changed.  Thus, for the example just cited, a new release would be created for `moduleB`.  This is modelled on how the Java plugin's `buildNeeded` and `buildDependents` tasks work (described [here](https://docs.gradle.org/current/userguide/multi_project_builds.html#sec:multiproject_build_and_test)).
--- a/docs/configuration/tasks.md
+++ b/docs/configuration/tasks.md
@@ -1,13 +1,17 @@
 # Tasks
 
-`axion-gradle-plugin` adds 6 new Gradle tasks:
+`axion-gradle-plugin` adds 8* new Gradle tasks:
 
 -   *verifyRelease*
 -   *release*
 -   *createRelease*
+-   *releaseDependents*
+-   *createReleaseDependents*
 -   *pushRelease*
 -   *currentVersion*
 -   *markNextVersion*
+
+*The plugin also creates a helper task called `configureReleaseDependentsTasks` that is called internally, but this is not designed to be called by users.
 
 ## verifyRelease
 
@@ -28,6 +32,14 @@ of these tasks running exactly one after another.
 ## createRelease
 
 Run pre-release actions ([Pre/post release hooks](hooks.md)) and create release tag.
+
+## releaseDependents
+
+Works as per `release`, except that releases are created on any submodules that depend on the project in which this task is called, including transitive dependencies (i.e. changes to `moduleA` will generate releases for `moduleB` and `moduleC`, assuming `moduleC` depends on `moduleB` depends on `moduleA`).  This is described in further detail in [Basic Usage](../basic_usage.md).
+
+## createReleaseDependents
+
+Works as per `createRelease`, except that releases are created on any submodules that depend on the project in which this task is called, including transitive dependencies (i.e. changes to `moduleA` will generate releases for `moduleB` and `moduleC`, assuming `moduleC` depends on `moduleB` depends on `moduleA`).  This is described in further detail in [Basic Usage](../basic_usage.md).
 
 ## pushRelease
 
--- a/src/integration/groovy/pl/allegro/tech/build/axion/release/MultiModuleProjectDependencyIntegrationTest.groovy
+++ b/src/integration/groovy/pl/allegro/tech/build/axion/release/MultiModuleProjectDependencyIntegrationTest.groovy
@@ -1,0 +1,407 @@
+package pl.allegro.tech.build.axion.release
+
+import org.gradle.testkit.runner.TaskOutcome
+
+import java.util.stream.Collectors
+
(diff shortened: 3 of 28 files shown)


john-tipper:
Coveralls seems to be complaining about lack of tests in `DummyRepository`, which is a class that does nothing and doesn't seem sensible to test.  Please let me know what you think, happy to edit as required.

john-tipper:
I have found that when there are large numbers of submodules (50+) and Gradle builds in parallel, the creation of the release tags causes contention on the Git repo and some of these then randomly fail without an error message.  The release tag is then rolled back and the push command then fails because the object ID of the commits no longer exists.  The exception that is caused looks like:

```
Caused by: java.io.IOException: Source ref refs/tags/example-0.1.3 doesn't resolve to any object.
--
160 | at org.eclipse.jgit.transport.RemoteRefUpdate.<init>(RemoteRefUpdate.java:298)
161 | at org.eclipse.jgit.transport.RemoteRefUpdate.<init>(RemoteRefUpdate.java:199)
162 | at org.eclipse.jgit.transport.Transport.findRemoteRefUpdatesFor(Transport.java:652)
163 | at org.eclipse.jgit.transport.Transport.findRemoteRefUpdatesFor(Transport.java:1469)
164 | at org.eclipse.jgit.api.PushCommand.call(PushCommand.java:167)
165 | ... 106 more

```

The solution is for the release tasks to be run non-concurrently and this is achieved by having all release tasks declare a common directory as a task output.  Gradle ensures that only one release task executes at any one time.  Whilst it would theoretically be possible to use `mustRunAfter` to achieve this non-concurrent behaviour, in practice with a large project with potentially many interrelated dependencies I think it would be very difficult to do so in a manner that would guarantee no deadlocks and thus this was the easiest solution.

TabraizChel:
Was this ever merged in ?

bgalek:
@TabraizChel nope, as you see this branch is a bit stale and with conflicts

TabraizChel:
@bgalek yup granted this branch is stale but was another PR opened up at some point to add any similar kind of feature. 
I have the same issue mentioned in https://github.com/allegro/axion-release-plugin/issues/325 and if it's not been resolved since then in a different PR I'll update this PR

gwweaver:
Is there any other way to accomplish a similar functionality to this using the Axion plugin? @john-tipper  Or perhaps even changing where Axion detects changes for the release?

shashken:
I am not sure I understood everything here perfectly, I will share my use case & proposal and see if it fits,
We are moving away to use composite build, as a result when a dependency project gets updated, the main project stays on the same version, this created all sorts of problems for us so we reverted, I opened a new PR #647 that is actually continuing the work I did in #422 to make monorepos work (closed #332)
They idea is to look for latest tag in both project dir and dependency dirs and to check for diff in commits in the deps dir as well
This allows a more direct (composite build) dependency mechanism in monorepos and can help with some use cases

bgalek:
we can get back to it if needed

PR #345, opened 2020-07-17
Added GitHub Actions release process

gavlyukovskiy:
Hi, I have spent some time to make things work with GitHub Actions. I though it will be helpful having it in the docs.

Example of fully configured flow: https://github.com/gavlyukovskiy/axion-release-with-github-actions/blob/master/.github/workflows/release.yml
run: https://github.com/gavlyukovskiy/axion-release-with-github-actions/runs/883701050?check_suite_focus=true


adamdubiel:
Great, thanks a lot!

PR #349, opened 2020-07-29
Improved handling of lightweight tags

a-27m:
Hi @adamdubiel
When peeling a non-annotated, i.e. lightweight tag, object ID can be `null`, according to [jgit docs](https://download.eclipse.org/jgit/site/5.8.1.202007141445-r/apidocs/org/eclipse/jgit/lib/Ref.html#getPeeledObjectId--). This patch compares ref to `HEAD` directly for such tags.
Let me know what you think!


adamdubiel:
Looks great :) Thanks!

PR #377, opened 2021-03-05
set default prefix and separator to v instead of release-

jandersson-svt:
Background:

There are two defacto-standards for tagging versions, vX.Y.Z and X.Y.Z.

Problem:

Anedoctal real life experience from many projects is that they often use the default
tagging prefix-configuration (currently release-)

What does this patch do?:

This patch would change the default release-X.Y.Z tagging to vX.Y.Z.
No logic changes, it is basically a renaming patch.
EDIT: After discussion there are some small logic changes, to handle old vs new projects.

This would just be a more "defacto-standard" default prefix-configuration.

Note:

SemVer-specification does not specify prefixes but gives an example,
, https://semver.org/#is-v123-a-semantic-version

Signed-off-by: Josef Andersson <josef.andersson@svt.se>


bgalek:
Hi! @jandersson-svt! Thanks for your work! 
I love the idea to change `release` to `v`, I'll need to think about this PR a bit - maybe we should have a deprecation warning before doing a change like that? also, we need to make sure, that this name does not affect old repositories - i.e. the next release for release-1.0, release-2.0 won't be v1.0 but v3.0 :)). Maybe we should provide an easy way to "stay on the old road"? Any thoughts? :)

jandersson-svt:
Thanks for the quick feedback, and glad you liked the idea.

After thinking I have two suggestions on how one can solve it.

**Suggestion 1:** By documentation and major upgrade. - "The easy road for the dev, but might anger a few users that misses this"
Deperacting "release-" and bumping the major version (which allows) for non-backward compatibilty. Strongly suggst that anyone using release- now add an configurations for scm  a la scmVersion {
    tag {

        prefix = 'release' 
        versionSeparator = '--
    }
}
meh... angy users, possibly.

**Suggestions 2** - A little bit of more work dev, seamless for users.

func PseudocodeSetPrefixAndSep {

hasExistingPrefixAndSeparator tags as release- - set them like that now too (release-)
has no preexisting "release-"tag history - use the new default  v
}
should be seamless for users, old projects as well as new.

External conf always overrides, just like it is working now.

I would prefer solution 2 here. What do you think?

Anyway, I was also thinking of using just two global/static conf values for Prefix and Separator or a little miniconfobject holding these , makes it easy with refactors etc..Besides It would also make solution suggestion 2 above easier to get done also I guess.. I'll get too this at once and I'll update the PR with a suggestion for this. 



//Does the project already have tags that all contain release-?  If so use release- as default values..
Otherwise

bgalek:
Suggestion 2 - I was thinking exactly about this!
Feel free to update the PR, we'll discuss technical details during code review ;)

jandersson-svt:
Great! Let's go for Suggstion 2 then:)

Commit 2: I replaced all literal strings of prefix and separator, and fixed so they are only set in one place from a conf-object. Makes it easy to run all tests with different prefix-separator values. Tested with v and release-.
Great testing work in the project from you btw!!

So, Next thing to do for this PR is the small logic to ListAllTags->check if they have prefix release->if so, it's i an old project and should have relase- as defaults, otherwise leave as-is.. I think I know how to solve that and where, but need to check and try with real project also. Hopefully I'll get some time later for that.

jandersson-svt:
Added: check the tag history, and if found to be consisting of release- (and no conf set), that standard is used.

Added tests to GitRepository. Also test IRL locally.

Surely, You might want to move the few added functions if think they belong somewhere, or/and refactior them so it fits your intended projects style better.
But, other than that this seems to work:)

mockitoguy:
+1 to supporting "v" prefix. It maps to Mockito projects and other projects I have been maintaining in the past.

+1 to making it configurable without changing the default (customer retention and trust).

Thanks!

bgalek:
@jandersson-svt we should separate this change to 2 PR's - doc's/readme/references (we could merge it just before next release) and the actual change ;) I'll do it and I'll prepare for a release ;)

jandersson-svt:
Sounds great, thank!

PR #379, opened 2021-03-19
Make use of Gradle task configuration avoidance APIs (fixes #289)

### LINKED ISSUE #289: Use configuration avoidance API when creating tasks
Tasks are currently created eagerly with `project.tasks.create(...)`.  They could instead be created using the recommend "configuration avoidance" API with `project.tasks.register(...)`.

https://docs.gradle.org/current/userguide/task_configuration_avoidance.html

f4lco:
Hi 👋  I propose the following patch for #289. Using `register(...)` can significantly speed up Gradle builds because Gradle then delays task creation (and configuration) to the point when actually needed instead of doing all the work upfront.
@adamdubiel


realdadfish:
LGTM

bgalek:
@f4lco thanks for your contribution!

f4lco:
@realdadfish @bgalek thanks, your welcome. Glad I could help a bit.

PR #380, opened 2021-03-19
WIP: Version pinning

### LINKED ISSUE #328: Pin versions
My use case might sound a little bizarre at first, but hear me on. I have a Gitlab CI pipeline that operates in different stages. All these stages do not share the same workspace, but start with a blank one - and a freshly cloned Git repo - by default. (Gitlab does not support shared workspaces.)

My workflow nonetheless should / must look like the following:

* `stage1`: createRelease
* `stage2..n-1`: use release version in build
* `stagen`: pushRelease

As of now this is not possible, because I'd have not only to create the tag, but push it already in `stage1` so it becomes available in `stage2..n-1`, but I'd also would have to remove the tag from the remote repo again in case `stage2..n-1` fails. Finally, `stagen` becomes superfluous because the tag had to be pushed to the repo already anyways.

So, what I came up with was a custom Gradle Plugin that overrides / uses the Axion infrastructure to do the following:

*  have a new `pinVersion` task that writes out the output of `scmVersion.version` to a configurable file
* transport this file through the different stages
* have a custom `myPlugin.version` method that either reads out the file (if it exists) or proxies to `scmVersion.version` and a custom `OutputCurrentVersion` task that does the same
* have a custom `CreateRelease` that does the same

Now while step 1 to 3 was kind of easy to implement, step 4 wasn't. The problem is that the `VersionConfig` instance has no interface it implements (so I could write an easy proxy implementation) and both cglib and javassist threw errors because they couldn't cope with the fact that the actual instances of the class was `VersionConfig_Decorated` (thanks Gradle) and not `VersionConfig`.

Anyways, I had to resort to the ugliest implementation, basically proxying all methods in `VersionConfig` myself and only overriding the needed `VersionConfig.getVersion()` method that is also used by the `CreateReleaseTask`.

Now this all works, but it is a great hack. I wonder if there would be the possibility to make this easier, either by implementing something like "pinned versions" directly into the plugin itself or at least by making it easier to override the contents of the `VersionConfig` structure for "bizarre" use cases like mine :)

What do you think?

realdadfish:
Implement version pinning in Axion (closes #328)

TODO: 
* Missing unit tests for `VersionService`
* More integration tests

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -87,6 +87,8 @@ dependencies {
 
     implementation group: 'com.github.zafarkhaja', name: 'java-semver', version: '0.9.0'
 
+    implementation group: 'com.squareup.moshi', name: 'moshi', version: '1.11.0'
+
     testCompile (group: 'org.ajoberstar.grgit', name: 'grgit-core', version: '4.0.1') {
         exclude group: 'org.eclipse.jgit', module: 'org.eclipse.jgit.ui'
         exclude group: 'org.eclipse.jgit', module: 'org.eclipse.jgit'
--- a/src/integration/groovy/pl/allegro/tech/build/axion/release/BaseIntegrationTest.groovy
+++ b/src/integration/groovy/pl/allegro/tech/build/axion/release/BaseIntegrationTest.groovy
@@ -3,7 +3,7 @@ package pl.allegro.tech.build.axion.release
 import org.gradle.testkit.runner.BuildResult
 import org.gradle.testkit.runner.GradleRunner
 
-class BaseIntegrationTest extends RepositoryBasedTest {
+abstract class BaseIntegrationTest extends RepositoryBasedTest {
 
     File buildFile() {
         return temporaryFolder.newFile('build.gradle')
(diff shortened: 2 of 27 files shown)


realdadfish:
Updated to recent Axion master.

bgalek:
@realdadfish please also add docs info

realdadfish:
@bgalek Will do, if this PR actually has a chance of getting merged :)

Would you please make a code review before, so in case some things change (e.g. the enable flag should be removed, the default filename should be changed, ...) I don't document things twice. Thanks!

realdadfish:
@bgalek Friendly reminder :)

src/integration/groovy/pl/allegro/tech/build/axion/release/RemoteRejectionTest.groovy
```diff
@@ -21,6 +22,7 @@ import java.nio.file.Paths
 import static pl.allegro.tech.build.axion.release.TagPrefixConf.fullPrefix
 
 @Testcontainers
+@Ignore
```
bgalek, line 25: don't forget to unignore it before merging ;)

bgalek:
@realdadfish I'm not really comfortable with this plugin writing and reading files from the filesystem. That should be externalized - if somebody wants to use a DB, git-specific-feature, or anything like that.

Can't some kind of hooks solve your use case?
STAGE1 - onNewRealese(context) hook which you can extract the version from and save it wherever you like (you decide if it is a file, a DB or maybe your pipeline metadata storage)
STAGE2 - beforeRelease() hook, that you implement to check if version should be updated from outside and tagged again before going to the next stage

This way would be more "generic" than pinning feature. WDYT?

PR #384, opened 2021-03-25
Documented recent features

mockitoguy:
- documented `previousRevision` property
 - added a concise doc for the empty `tagPrefix` property
 - removed unnecessary spaces (happened automatically, I think it's OK to keep in this PR)


mockitoguy:
@bgalek, ok?

PR #410, opened 2021-07-07
Bump Gradle to 7.1.1

kkocel:
Simplified configuration - use testcontainers instead of gradle tasks to
setup docker test.

Bump groovy to 3.x (implicitly by bumping gradle)
Bump spock to 2.x, use `@TempDir` instead of Junit4 `TemporaryFolder` rule

Explicit Gradle version and distribution type in build.gradle

### DIFF
--- a/src/integration/groovy/pl/allegro/tech/build/axion/release/ExtendingTasksIntegrationTest.groovy
+++ b/src/integration/groovy/pl/allegro/tech/build/axion/release/ExtendingTasksIntegrationTest.groovy
@@ -34,7 +34,7 @@ class ExtendingTasksIntegrationTest extends BaseIntegrationTest {
         task << [
             'CreateReleaseTask',
             'MarkNextVersionTask',
-            'OutputCurrentVersionTask',
+            // 'OutputCurrentVersionTask', FIXME causes test to fail - Type 'pl.allegro.tech.build.axion.release.OutputCurrentVersionTask' property 'versionConfig.versionIncrementer.creator' is missing an input or output annotation
             'PushReleaseTask',
             'ReleaseTask',
             'VerifyReleaseTask'
--- a/src/integration/groovy/pl/allegro/tech/build/axion/release/ExtendingTasksIntegrationTest.groovy
+++ b/src/integration/groovy/pl/allegro/tech/build/axion/release/ExtendingTasksIntegrationTest.groovy
@@ -34,7 +34,7 @@ class ExtendingTasksIntegrationTest extends BaseIntegrationTest {
         task << [
             'CreateReleaseTask',
             'MarkNextVersionTask',
-            // 'OutputCurrentVersionTask', FIXME causes test to fail - Type 'pl.allegro.tech.build.axion.release.OutputCurrentVersionTask' property 'versionConfig.versionIncrementer.creator' is missing an input or output annotation
+            'OutputCurrentVersionTask',
             'PushReleaseTask',
             'ReleaseTask',
             'VerifyReleaseTask'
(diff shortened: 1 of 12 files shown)


src/integration/groovy/pl/allegro/tech/build/axion/release/ExtendingTasksIntegrationTest.groovy
```diff
@@ -34,7 +34,7 @@ class ExtendingTasksIntegrationTest extends BaseIntegrationTest {
         task << [
             'CreateReleaseTask',
             'MarkNextVersionTask',
-            'OutputCurrentVersionTask',
+            // 'OutputCurrentVersionTask', FIXME causes test to fail - Type 'pl.allegro.tech.build.axion.release.OutputCurrentVersionTask' property 'versionConfig.versionIncrementer.creator' is missing an input or output annotation
```
bgalek, line 37: what about that? :)

src/main/java/pl/allegro/tech/build/axion/release/domain/hooks/CommitHookAction.java
```diff
@@ -24,7 +24,7 @@ public String call(Object... args) {
 
     @Override
     public void act(HookContext hookContext) {
-        String message = (customAction.call(hookContext.getCurrentVersion(), hookContext.getPosition())).toString();
+        String message = (customAction.call(hookContext.getReleaseVersion(), hookContext.getPosition())).toString();
```
bgalek, line 27: typo?

bgalek (changes requested):

src/main/java/pl/allegro/tech/build/axion/release/domain/hooks/CommitHookAction.java
kkocel: I saw it's deprecated and `getReleaseVersion()` uses the same underlying private field:

```java
    public String getReleaseVersion() {
        return currentVersion;
    }

    @Deprecated
    public String getCurrentVersion() {
        return currentVersion;
    }
```

Should I revert this or leave at it is?

src/integration/groovy/pl/allegro/tech/build/axion/release/ExtendingTasksIntegrationTest.groovy
kkocel: As I dmed you - this is the test that uncovers some missing Gradle task annotations (`@Input`, `@Output` etc.). 
I am not familiar with all the plugin domain logic so I left this out intentionally.
I saw that you added those annotations to other parts of the code so would you mind looking into this and adding appropriate annotations?

as stated in the error message - `creator` is not annotated but it's a local variable:

```groovy
    static Closure<Version> versionIncrementerFor(String name, Map configuration = [:]) {
        PredefinedVersionIncrementer creator = values().find { it.name == name }
        if (creator == null) {
            throw new IllegalArgumentException("There is no predefined version incrementer with $name name. " +
                "You can choose from: ${values().collect { it.name }}")
        }
        return { VersionIncrementerContext context -> creator.versionIncrementer(context, configuration) }
    }
```

kkocel: Ok, I got away with simpler solution - instead of calling `versionIncrementerFor` I used direct enum value

src/main/java/pl/allegro/tech/build/axion/release/domain/hooks/CommitHookAction.java
bgalek: no, it's ok

bgalek (approved):

bgalek (approved):

bgalek:
LGTM, thx for contribution! @kkocel

PR #411, opened 2021-07-10
Fix path of ScmPosition class source file in hooks.md

rzabini:


bgalek:
@rzabini thx for the fix! :)

PR #412, opened 2021-07-10
Custom package prefix for relocated jars

kkocel:


bgalek:
@kkocel I had a problem with this setting - it seemed that some part of jgit was loading classes by their fully-qualified name, did you test that setting with multiple repository authentication methods/protocols?

kkocel:
@bgalek Please note that JARs are _already_ repackaged.

This is decompiled output from 1.13.3 release:
Screenshot 2021-07-12 at 08 48 31

as you can see jgit is `shadow` prefixed.

The goal of this PR is to avoid conflict with other shadowed plugins ( see: https://github.com/n0mer/gradle-git-properties/pull/184 ) by introducing fairly unique prefix, other than `shadow`

bgalek:
@kkocel ok, thx for explanation

bgalek (approved):

Issue #332, opened 2020-04-25
CurrentVersion is immediately SNAPSHOT after release


JozoVilcek:
Hi, I have a git repository, monorepo, hosting many independent projects. Each projects pulls in `axion-release-plugin` configures is prefixTag and does it's own release cycles.

The release process is essentialy 1.) check the build, 2) release, 3) build deployable artifact for released version. This process relies on the fact that if `cV` is e.g. `1.0.1-SNAPSHOT` then right after release it will be `1.0.1`, therefore, without snapshot.

This works fine up to axion v`1.10.3`. In v`1.11.0` it breaks. After release, the `cV` for the above gives `1.0.2-SNAPSHOT`. I think the change is introduced here: https://github.com/allegro/axion-release-plugin/commit/7202376bba5549a2ac75976f2bf4730638c02f25

In `VersionResolver` the `ScmPosition position = repository.currentPosition()` is changed to 
`latestChangePosition = repository.positionOfLastChangeIn(projectRootRelativePath, versionRules.monorepoProperties.dirsToExclude)`

In case of multiple coexisting projects, it is very likely to not to be on the last change for project for which release is being done (something merged to master between merge of a project and invoking a release). Then `boolean onCommitWithLatestChange = currentVersionInfo.commit == latestChangePosition.revision` is going to be `false` and this will set `VersionInfo.onReleaseTag` to `false` and leads to `isSnapshot`.

If my conclusion is correct, then how do I get the correct release version (non snapshot) right after performing a release?

sradi:
I am having that issue, too.

sradi:
After short analysis it seems, that the attribute `projectDirs` of the monorepos configuration is empty, when the currentVersion task is executed.
If it was set correctly, the `lastChangePosition` would probably be calculated correctly, so that the resulting value of `onCommitWithLatestChange` would  be `true`, as expected.

sradi:
@john-tipper would you mind, having a look?

adamdubiel:
I suspect that this could have something to do with the way plugin is initialized. If `dirsToExclude` is set using some gradle `project.projectDir` and is empty for `currentVersion`, then it might be because of how `scmVersion` being eager (not lazy).

I know that Gradle introduced mechanisms for lazy loading of properties some time ago, meaning `project.version = scmVersion.version` could be evaluated later in the cycle, which probably would fix a lot of issues.

stefan-rademacher-tk:
[...]

PR #422, opened 2021-10-06
Add support for jgit log search for latest commit when path is given.

shashken:
Add support for jgit log search for latest commit when path is given.
Solved an edge case where the latest found was actually one commit prior to the latest commit (the merge to master)
Tries to solve #332 
@bgalek - please review
@farrukhnajmi FYI


shashken:
hmm I thought of something, I think the whole approach for monorepo with excludeDirs was wrong to begin with,
Take this example case:
A (Head)
 |
B (Tag 1.2.0)
 |
C
|
D
|    
E  (latest change found for path)
|
F (no change to path)
|
G (Tag 1.1.0)

what we expect: 
A (1.2.0)
B-E (1.1.0-snapshot)
F-G (1.1.0)

what is currently happening:
A (1.2.0-snapshot)
B (1.2.0-snapshot)
C-G( 1.1.0-snapshot)

This happens because we check if latestChange == tag, but we dont have to use this approach, instead we can go 
Diffformatter.setPathFilter(path);
Diffformatter.scan(latestChange, taggedCommit)

bgalek:
I'll try to look at it today

Issue #431, opened 2021-11-17
Github Actions support


big-andy-coates:
I've been setting up a new project using Github's own build Actions and it took a little bit of digging to get the release plugin to work.  Thought I'd share the magic incantation needed...

Obviously, the plugin needs the tags to be available so it knows what the next tag should be.  By default, the `actions/checkout@v2` action doesn't include tags and is only a shallow checkout.

I found adding `git fetch --tags --unshallow` prior to doing a `release` was all that was needed to get everything working, e.g. https://github.com/creek-service/creek-core/blob/main/.github/workflows/gradle.yml#L33

Not sure if you want to add this to your docs?

gabrieljones:
`git fetch --tags --unshallow` should be added to the documentation here:
https://axion-release-plugin.readthedocs.io/en/latest/configuration/ci_servers/#github-actions

bgalek:
@gabrieljones just use
```
scmVersion {
    unshallowRepoOnCI.set(true)
}
```

it will be turned on in upcoming releases ;)

bgalek:
@gabrieljones I thnk not, since we don't have a generic way of introducing those flags, should be easy to add this, I'm ok with that - maybe you would like to create a PR? If not, create an feature request for that and I'll be happy to add this.

PR #432, opened 2021-11-18
#431 Doc config for Github Actions

big-andy-coates:
Fixes: https://github.com/allegro/axion-release-plugin/issues/431

Add docs covering how to get the plugin to work with Github's default checkout action.

### DIFF
--- a/docs/configuration/github_actions.md
+++ b/docs/configuration/github_actions.md
@@ -1,0 +1,46 @@
+# Using plugin with GitHub Actions
+
+The plugin can be used with builds running via GitHub Actions.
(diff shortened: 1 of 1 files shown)


bgalek:
Hi! Is there any difference to:
```
      - uses: actions/checkout@v2
        with:
          fetch-depth: 0 
```
?

big-andy-coates:
> Hi! Is there any difference to:
> 
> ```
>       - uses: actions/checkout@v2
>         with:
>           fetch-depth: 0 
> ```
> 
> ?

Both should work, however,  the above creates a full deep copy of the repo locally even for PR builds, which will slow down builds. Where as the example I've added uses a shallow clone of the repo for PR builds and upgrades it to a full copy for release builds.

PR #433, opened 2021-11-24
docs: add gitlab ci example

janderssonse:
Signed-off-by: Josef Andersson <josef.andersson@gmail.com>


janderssonse:
Adds a ci example for gitlab to the docs. There was non before, and this is how I run it currently at least, so it is "tested in prod"

bgalek:
nice, thx!

Issue #443, opened 2022-02-09
Simplify usage on Gitlab CI with versionWithBranch


david0:
Gitlab CI seems to do a detached HEAD checkout (no branch name set). I did not find a way to tell not to do so.

We are using the `versionWithBranch` versionCreator and on feature branches it will generate the same version as on the main/master branch.

I was able to work around that by using `-Prelease.overriddenBranchName=$CI_COMMIT_BRANCH`. Maybe the plugin could be improved to take CI_COMMIT_BRANCH into account? Or, at least this workaround should be noted in the Gitlab CI section of the documentation.

mklueh:
I was wondering why I could not get snapshots using Gitlab CI when on my dev branch. When will this be rolled out?

bgalek:
https://axion-release-plugin.readthedocs.io/en/latest/configuration/ci_servers/#gitlab-ci

PR #445, opened 2022-02-22
Document how to override the branch name on Gitlab CI

david0:
Improves #443.

I did not test releasing yet, but I assume it is not really connected to this setting.


bgalek:
thx!

PR #446, opened 2022-03-01
Test reports as artifacts

kkocel:


bgalek:
hi @kkocel, why do we need to persist test result artifact?

kkocel:
there is no need, only possiblity :) 
having access to reports you are able to see the whole stacktrace and standard output (which is not possible with test-report plugin)

kkocel:
if you don't think it's needed feel free to close/reject the PR :)

bgalek:
maybe gradle --scan would be better than?

PR #460, opened 2022-05-19
shadow gradle plugin removed

bgalek:


build.gradle
```diff
@@ -91,26 +88,10 @@ dependencies {
...
-    prefix = "axion"
-}
-
-tasks.shadowJar.dependsOn tasks.relocateShadowJar
-
-shadowJar {
-    archiveClassifier.set('')
-}
-
 components.java.withVariantsFromConfiguration(configurations.runtimeElements) {
```
cloudshiftchris, line 95: This section is only for shadow, removing dependencies from POM for publishing (as they are included in the shadow jar).  Should be removed to publish a full POM with dependencies.

build.gradle
```diff
@@ -57,8 +54,8 @@ sourceSets {
 }
 
 dependencies {
-    shadow gradleApi()
-    shadow localGroovy()
+    implementation gradleApi()
```
cloudshiftchris, line 57: gradleApi() is automatically added by java-gradle-plugin, and not required here.  If we do with to leave it, it should be 'compileOnly' as the target gradle install (where the plugin is applied) provides this.

build.gradle
```diff
@@ -57,8 +54,8 @@ sourceSets {
...
+    implementation localGroovy()
```
cloudshiftchris, line 58: localGroovy() should be in compileOnly, as Groovy is provided by Gradle as the plugin is applied.

build.gradle
bgalek: yes, I've just tested this pr and found this out ;)

build.gradle
bgalek: thanks for pointing this out!

Issue #481, opened 2022-06-02
Broken release on Maven Central


wiget:
Jar for version 1.13.10 released to Maven Central is broken in some way. 
```
* What went wrong:
Could not apply requested plugin [id: 'pl.allegro.tech.build.axion-release', version: '1.13.10'] as it does not provide a plugin with id 'pl.allegro.tech.build.axion-release'. This is caused by an incorrect plugin implementation. Please contact the plugin author(s).
> Plugin with id 'pl.allegro.tech.build.axion-release' not found.
```
Jar from Gradle Plugins works without problem.

https://plugins.gradle.org/m2/pl/allegro/tech/build/axion-release-plugin/1.13.10/axion-release-plugin-1.13.10.jar has md5 hash `fb245869dbbf1ebfd0eb25fe83c128fa`  
but https://repo1.maven.org/maven2/pl/allegro/tech/build/axion-release-plugin/1.13.10/axion-release-plugin-1.13.10.jar has md5 hash `d39481b8f5dc729250d2086ae313ba72`

bgalek:
@wiget thx for pointing out, gonna check it out as soon as possible!

bgalek:
I'm trying to figure out the problem, the faster way will be to release next version, @wiget will you check it?:)

wiget:
Sure, I can test. Just let me know version number.

bgalek:
@wiget i've released v1.13.11 - please check it out - this project is not ready to use anykind of gradle plugin snapshots ;(

wiget:
@bgalek comparing your last commit to Gradle [documentation](https://plugins.gradle.org/docs/publish-plugin#example) I think `gradlePlugin` should be defined outside `pluginsBundle`

Something like this
```
gradlePlugin {
    testSourceSets project.sourceSets.integration
    setAutomatedPublishing(false)
    plugins {
        release {
            id = 'pl.allegro.tech.build.axion-release'
            implementationClass = 'pl.allegro.tech.build.axion.release.ReleasePlugin'
        }
    }
}

pluginBundle {
    website = 'https://github.com/allegro/axion-release-plugin'
    vcsUrl = 'https://github.com/allegro/axion-release-plugin'
    description = 'Release and version management plugin.'
    tags = ['release', 'version']
    plugins {
        release {
            displayName = 'axion-release-plugin'
        }
    }
}

```

bgalek:
[...]

PR #487, opened 2022-06-06
fix: customize pom, not create another publication

wiget:
Related to #481


bgalek:
When publishing a snapshot to maven central (or local) I can see 
pl.allegro.tech.build.axion-release.gradle.plugin pom beeing published

PR #493, opened 2022-06-09
GH-490 Support encoding in `fileUpdate` hook (Fix #490)

### LINKED ISSUE #490: `fileUpdate` hook does not respect encoding
Hooks applied to files (like e.g. README) with e.g. emojis replace content with question marks:

obraz

I also don't see any possibility to change the encoding in configuration:

* https://axion-release-plugin.readthedocs.io/en/latest/configuration/overview/

dzikoysk:


bgalek:
@dzikoysk LGTM, we only need to update docs :)

PR #578, opened 2023-02-08
GitHub dependabot friendly dependency versions

big-andy-coates:
Looks like GitHub's dependabot is configured for this repo, but it's not picking up a few dependency versions because of unusual Kotlin structures in the build files used to store the version. This PR switches to a more standard / dependabot-friendly way of defining dependency versions.

(Dependabot uses simplistic code parsing to detect dependency versions. It can't handle funky Kotlin stuff)

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -35,12 +35,6 @@ repositories {
     }
 }
 
-object Versions {
-    const val jgit = "5.13.1.202206130422-r"
-    const val jsch = "0.1.55"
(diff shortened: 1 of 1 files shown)


bgalek (approved):

bgalek:
thx! meant to do that a log time ago;)

big-andy-coates:
@bgalek fancy manually kicking off dependabot ?

bgalek:
Done :)

PR #589, opened 2023-02-28
readme update

sreich:
add link to plugin gradle page. also fixes the version on the badge. 

unfortunately auto grabbing that version doesn't seem to work so we need to be explicit about it and update in the future still...better than before which was showing a very wrong version...


bgalek (approved):

bgalek:
sure, thx!

PR #610, opened 2023-04-13
Switching to jsch implementation by mwiede

rzukow:
Hi! 

I discovered that authorization fails with invalid private key error when private key is generated by a new versions of openssh (header that is not supported is : -----BEGIN OPENSSH PRIVATE KEY-----)

The problem is that jsch is not actively maintained, but there is a for, which is.
See: https://stackoverflow.com/questions/73862879/workaround-for-jsch-unsupported-kex-algorithms

After switching to new implementation, problem is solved. Private key is properly used.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -62,13 +62,26 @@ dependencies {
     implementation("org.eclipse.jgit:org.eclipse.jgit:$jgitVersion")
     implementation("org.eclipse.jgit:org.eclipse.jgit.ssh.jsch:$jgitVersion")
     implementation("com.github.mwiede:jsch:$jschVersion")
-    implementation("com.jcraft:jsch.agentproxy.core:$jschAgentVersion")
-    implementation("com.jcraft:jsch.agentproxy.jsch:$jschAgentVersion")
-    implementation("com.jcraft:jsch.agentproxy.sshagent:$jschAgentVersion")
-    implementation("com.jcraft:jsch.agentproxy.pageant:$jschAgentVersion")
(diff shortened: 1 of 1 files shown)


bgalek:
Hi @rzukow! That's cool!
Unfortunately, build is failing, wanna check it out?

bgalek:
from readme:
> Addition: You can further exclude any of com.jcraft:jsch.agentproxy.jsch, com.jcraft:jsch.agentproxy.core or com.jcraft:jsch.agentproxy.pageant, because these modules where integrated in this fork (see release notes of [0.1.66](https://github.com/mwiede/jsch/releases/tag/jsch-0.1.66)).

rzukow:
> Hi @rzukow! That's cool! Unfortunately, the build is failing; wanna check it out?

yeah, I can check that later today. Do you have a good place to add integration tests testing authentication with new style keys?

bgalek:
you can do manual test: https://github.com/allegro/axion-release-example
or use the dockerized tests in this repo

rzukow:
I will close it for now - I'm not sure if I'm able to fix tests. When I have a working solution I will make a PR again. I was hoping this is a one liner.

Here is what I know: replacing original jsch dependency with the one provided by mwiede is working fine and is handling more formats of keys for authorization to github. I tested it manually.

Unfortunately this library is multi release jar, and it does not work with this version of Gradle or Gradle Test Kit  or Spock and tests are failing

When I upgraded to Gradle 6.9.4 and changed toolchain to java 16 it kind of looks promising - I have a majority of integration tests working. But it requires some time to fix remaining tests...

Upgrading to Gradle 8 is a bit harder and I failed with this one.

Issue #622, opened 2023-05-19
VersionProperties.Incrementer in Kotlin != groovy.lang.Closure


duschata:
Hallo All,

working with the just fixed #590 issue. I run into the next problem using kotlin dsl.

Have a look at the failing test here:

https://github.com/duschata/axion-release-plugin/blob/mcve_kotlin_closure/src/integration/groovy/pl/allegro/tech/build/axion/release/KotlinDslCompatibilityTests.groovy#L89

Kotlin needs a strong typed VersionProperties.Incrementer here, but the condition in the VersionPropertiesFactory

https://github.com/duschata/axion-release-plugin/blob/mcve_kotlin_closure/src/main/groovy/pl/allegro/tech/build/axion/release/infrastructure/config/VersionPropertiesFactory.groovy#L68-L74

evaluates a (!)Closure and the kotlin lambda runs always into the wrong condition.

What can we do there, 

* using a kind of closureOf function in the kotlin script?
* fix the (!(value instanceof Closure))?

what do you think?

kind regards, Tom

bgalek:
Hi @duschata, nice work finding the issue. 
I would aim for fixing the `instanceof` but dunno if it won't be a big rewrite, wanna try to fix it?

duschata:
Hi @bgalek ,
I fixed it, wrote test & docs and it seems to work now. I've sent a PR,
greetings Tom

duschata:
Hi All, 
ups, there is nearly the same issue with VersionProperties.Creator. I'll try to fix this soon...

PR #627, opened 2023-05-26
#622  fix VersionVersionProperties.Creator

duschata:
instanceof VersionVersionProperties.Creator is also evaluated in the VersionPropertiesFactory using kotlin.dsl. Tests & Docu added


bgalek:
@duschata perfect, thx!

PR #629, opened 2023-05-30
update next_version.md with correct default strategy

agavra:
it seems the default behavior is incrementPatch, not incrementMinor - this PR updates the documentation to indicate that (this is in line with the docs on https://axion-release-plugin.readthedocs.io/en/latest/configuration/version/)


agavra:
Ah I actually think I misunderstood the difference between markNextVersion and release default behavior.

Issue #561, opened 2022-11-30
Regressions when upgrading from 1.13.6 to 1.14.3


realdadfish:
I'm testing with Gradle 7.5.1 and 7.6 and see several regressions after the update. This is my configuration:

```
configure<VersionConfig> {
        localOnly.set(true)

        repository {
            // I'm in a multi-component build, this wraps the search for a parent directory containing a .git subdir 
            directory.set(layout.dir(provider { getGitRoot() }))
        }

        tag {
            prefix.set("my-release-")
            versionSeparator.set("")
        }

        versionCreator(PredefinedVersionCreator.VERSION_WITH_BRANCH.versionCreator)
        branchVersionCreator.putAll(
            mapOf(
                // do not add branch names to versions of release and nightly builds
                "release/.*" to "simple",
                "develop" to "simple"
            )
        )

        snapshotCreator { _, scmPosition ->
            "-dev.${scmPosition.shortRevision}"
        }

        checks {
            // Can't rely on that because of JGit (https://github.com/allegro/axion-release-plugin/issues/316)
            uncommittedChanges.set(false)
        }
}
```

I'm seeing two issues:

* SNAPSHOTs don't seem to be recognized anymore and versioning is off. If I have a `my-release-1.1.0` tag in the parent branch and on branch `foo` I previously got `1.1.1-foo-dev.abcd1234`, now I'm only getting `1.1.0-foo`
* More importantly, the execution time of the `currentVersion` task is enormous (several minutes), where the task previously executed within seconds. If I execute the build then with `--debug`, Gradle spits out tons of debug messages like `Waiting to acquire shared lock on daemon addresses registry.`. The internets tell me that this is probably because of a memory issue (HEAP too low, etc.), but nothing changed in this regard. My `gradle.properties` is unchanged:

```
org.gradle.caching=true
org.gradle.parallel=true
org.gradle.jvmargs=-Xmx8g -XX:MaxPermSize=8g -XX:+HeapDumpOnOutOfMemoryError -Dkotlin.daemon.jvm.options\="-Xmx2048M"
```

I can provide a build scan (before / after) if this is of any use.

bgalek:
@realdadfish I'm a bit worried about `task is enormous (several minutes)`, would you like to try and dig into the problem?

bgalek:
about the snapshot problem, it should be easy to add such unit test - it would be easier to fix it then ;)

mpern:
[...]

PR #640, opened 2023-07-27
remove @InputDirectory to avoid memory leak (#561)

mpern:
Confirmed [my suspicion](https://github.com/allegro/axion-release-plugin/issues/561#issuecomment-1653540205):

After changing `RepositoryConfig.directory` to `@Input Property<String>`, `./gradlew currentVersion` finishes immediately (again).


bgalek:
@mpern perfect! Thanks!

gabrieljones:
This was a breaking change and should have resulted in a major version bump.

bgalek:
@gabrieljones hi! Thanks for pointing it out, this was unintentional. What did you need to update with your build setup?

PR #647, opened 2023-08-15
Add dependency dirs for monorepo config

shashken:
Add the option to configure several dependency directories to affect the version in a monorepo project


shashken:
@bgalek WDYT

shashken:
@bgalek WDYT?

bgalek:
@shashken https://github.com/allegro/axion-release-plugin/actions/runs/6183866024/job/16786446050?pr=647

shashken:
thanks! @bgalek working on it, we are also currently working with java 8 on a single project, is it possible to release this for older java version as well as a hotfix? I saw that latest versions removed support for older java versions

bgalek:
@shashken we could backport, but I really don't like the idea, jdk8 is dead to me ;)

shashken:
lol @bgalek as it should be 😄  but sadly not for us just yet 😢 
I can handle the backport, can you point me out to the proper flow I need to make for it here?

bgalek:
> lol @bgalek as it should be 😄 but sadly not for us just yet 😢 I can handle the backport, can you point me out to the proper flow I need to make for it here?

I would recommend a fork then, I don't want to manage two versions at the same time (v2), and this is a new feature not fixing a bug, sorry

shashken:
Ok so this one is ready from my side
Do you think it will be possible to release a hotfix on this repo for 1.14.x?
It will be a bit challenging building a pipeline on our side just for 1 version
Will it be simple in this repo?

bgalek:
thank you for understanding and your collaboration!
It should be very basic - if you're sticking to 1.14.x you could make the change locally and update the resulting jar to your maven without any pipelines (YAGNI) - assuming that you will migrate to 1.15.x and get back to main upstream someday ;)

bgalek:
@shashken please also upadate documentation

bgalek:
@shashken nice!
Last thing, could you add at least one test using `monorepos` settings? 
also I think we could use "monorepo" instead of plural version

shashken:
@bgalek added a test, this branch was opened 1.5 months ago, I am loosing a lot of context here, lets aim for short living branches and move the rename to a different issue/PR
WDYT?

bgalek:
Sure, we can

bgalek (approved):

PR #686, opened 2023-12-03
remove VersionScmPosition as it is a duplicate of ScmPosition

rzabini:
Using `scmVersion.scmPosition` property to add information to a jar manifest, I noticed that the class `VersionScmPosition` was missing the `isClean` property.
I though about adding another field to it, but then I saw that this class is basically a replica of `ScmPosition` (which has the `isClean` field), so maybe it is cleaner to remove altogether the class, and use `ScmPosition` instead.
What do you think?


bgalek:
Hi, seems like a good idea! Care for a PR? :)

bgalek:
LGTM :)

gabrieljones:
please add this back and deprecate

bgalek:
hi @gabrieljones, which fields are you missing? we can add them to ScmPosition class

gabrieljones:
The return type change on `getScmPosition` is causing the exception. I created a pre compiled convention plugin that calls `getScmPosition`. Downstream users of this plugin that have upgraded their axion release version are seeing the `NoSuchMethodError`.

gabrieljones:
If I am the only one experiencing this then we can close this as won't fix.

Issue #633, opened 2023-06-14
Unable to use latest SSH features


wdroste:
Summary jsch is abandoned and BitBucket (and GitHub) recommends the latest SSH key type ex: ECDSA

We followed bitbucket's instructions and our builds started to fail because jsch couldn't use the latest ECDSA private key and therefore axion-release-plugin failed to tag our repository using Git over SSH.

For more information:
https://github.com/gradle/gradle/pull/20387

Drop in replacement:
https://github.com/mwiede/jsch

bgalek:
hi @wdroste some people already tried that, https://github.com/allegro/axion-release-plugin/pull/610/files and it seems it won't be so easy as "drop in" :(

bgalek:
Should be fixed

wdroste:
Awesome thanks

PR #711, opened 2024-02-11
Switch to a recent version of a supported Jsch fork

mc1arke:
The version of JSch bundled in the plugin has not been maintained in a few years so has not kept up-to-date with newer cryptographic keys, ciphers, and signatures. This causes challenges for users attempting to use Git with configuration using up-to-date cryptographic recommendations, as the library only supports SHA1 signatures for RSA which distributions like OpenSSH have stopped supporting.

The JSch implementation has been switched to an actively maintained fork and an up-to-date version of BouncyCastle included in the dependencies to ensure elliptic curves and recent cipher suites. The existing RemoteRejectionTest has been altered to use public key authentication and an up-to-date container running a recent version of OpenSSH to allow realistic connectivity testing of remote Git over SSH connections.

As the Jsch distribution is a multi-version JAR including Java 16 class files, the plugin has had to be updated to Gradle 7 as the ASM version used in Gradle 6 is not compatible with Java 16 classes.


mc1arke:
This should fix #633

bgalek:
Hi, thank you for this contribution!
I'll check it out

bgalek:
@mc1arke hi! I like the change, I'll merge this, test it out and release new minor version!

bgalek:
@mc1arke released, thx!

PR #737, opened 2024-04-19
output released version in github actions environement

bgalek:


aleksandrserbin (approved):

src/main/groovy/pl/allegro/tech/build/axion/release/OutputCurrentVersionTask.groovy
```diff
@@ -32,5 +32,14 @@ abstract class OutputCurrentVersionTask extends BaseAxionTask {
 
         OutputWriter output = new OutputWriter()
         output.println(outputContent)
+
+        if (isGitHubWorkflowContext()) {
+            def gitHubOutputFile = new File(System.getenv('GITHUB_OUTPUT'))
+            "current-version=${outputContent}\n" >> gitHubOutputFile
```
radoslaw-panuszewski, line 38: If we do it here, it will still require to explicitly run `./gradlew currentVersion` task to produce the step output with current version. I would prefer it to happen during every `./gradlew release` (only we would need to rename the output variable to `released-version`).

Alternatively, outputting the `current-version` seems like a very general (yet useful) feature that I would like to have on every `./gradlew` execution. I can think of it as a responsibility of the `setup-gradle` action, not any plugin (let actions do Github stuff and plugins do Gradle stuff). It could be implemented like this:
```kotlin
// init.gradle.kts

rootProject {
    afterEvaluate {
        File(System.getenv('GITHUB_OUTPUT')).writeText("current-version=$version)
    }
}
```

bgalek: We won't need to init.gradle in our case, so I think we should implement this feature here anyway - to be used in the future ;)

radoslaw-panuszewski: Implementing it here will require you to apply `axion-release` plugin if you want to use the reusable workflow `java-package`. IMO it should be done in our `setup-gradle` action so any other plugin for reading version from nearest tag (or even `git describe`) can be used- the `current-version` output will always be present in every step executing `./gradlew`

bgalek: yes, but I think this feature can be separated from our usecase

radoslaw-panuszewski: As discussed offline - let's do it as described in my first proposition:
>  I would prefer it to happen during every ./gradlew release (only we would need to rename the output variable to released-version).

So basically: move it to `release` task and rename from `current-version` to `released-version`

bgalek: @radoslaw-panuszewski code updated!

src/integration/groovy/pl/allegro/tech/build/axion/release/SimpleIntegrationTest.groovy
```diff
@@ -34,6 +39,24 @@ class SimpleIntegrationTest extends BaseIntegrationTest {
         result.task(":currentVersion").outcome == TaskOutcome.SUCCESS
     }
 
+    def "should define an github output when running release task in github workflow context"() {
```
mkosmul, line 42: an -> a

mkosmul (approved):

rafalh:
I think you should add some info about this feature in docs. I would not expect gradle plugin to mess with github variables under the hood so it may be trickier to analyse what is happening in CI for people who do not know about this change. Good docs would probably help.
I assume this is something people are supposed to use and not some kind of PoC (in the letter case having it undocumented would make sense)

bgalek:
Yes, I'm going to update docs

bgalek:
https://axion-release-plugin.readthedocs.io/en/latest/configuration/basic_usage/#github-workflow-context

Issue #736, opened 2024-04-19
Retrieving the version for large repositories is slow


balrok:
When following the recommended way to set the version in a gradle project:
```gradle
project.version = scmVersion.version
```
it can take a considerable amount of time to compute it for larger projects.

## How to reproduce (real project)

* Clone a bigger project: `git clone https://github.com/spring-projects/spring-framework.git` (28k commits, 10k files) (*note: this is actually not that big.*)
* Patch it to run `scmVersion.version`:
```diff
diff --git a/build.gradle b/build.gradle
index e8c3550a7c..6da12e0876 100644
--- a/build.gradle
+++ b/build.gradle
@@ -1,4 +1,5 @@
 plugins {
+      id 'pl.allegro.tech.build.axion-release' version '1.17.0'
 	id 'io.freefair.aspectj' version '8.3' apply false
 	// kotlinVersion is managed in gradle.properties
 	id 'org.jetbrains.kotlin.plugin.serialization' version "${kotlinVersion}" apply false
@@ -11,6 +12,10 @@ plugins {
 	id 'me.champeau.mrjar' version '0.1.1'
 }
 
+def start = System.currentTimeMillis();
+scmVersion.version
+println("Took " + ((System.currentTimeMillis() - start) / 1000) + "s")
+
 ext {
 	moduleProjects = subprojects.findAll { it.name.startsWith("spring-") }
 	javaProjects = subprojects.findAll { !it.name.startsWith("framework-") }
```
* spring-framework uses a version-scheme for its tags which are not like it is configured in the plugin. Since they do not seem to have an influence on the speed, simply delete them `git tag -d $(git tag)` or configure the plugin, that it works
* run `./gradlew cV` and watch out for the "Took Xs"

### Measurements:
* directly after cloning:
    * 0.5s - 2.0s
* adding 10k files: `for i in $(git ls-files|grep -v "/$"); do touch "$i.release-plugin-test"; done`
    * 1.5s - 3.5s
* adding 30k commits: `for i in $(seq 1 30000); do git commit --allow-empty -m"$i"; done`
    * 1.4s - 3.4s

## How to reproduce (real project)

Alternatively, one can create an empty git project with only build.gradle:
```gradle

plugins {
    id 'pl.allegro.tech.build.axion-release' version '1.17.0'
}

def start = System.currentTimeMillis();
scmVersion.version
println("Took " + ((System.currentTimeMillis() - start) / 1000) + "s")
```

[...]

PR #738, opened 2024-04-19
implement flag overridenIsClean

balrok:
This flag is implemented similar to overriddenBranch and allows to set the "isClean" flag for the repository. Which basically determines if the working tree is dirty or clean.

One use-case is to speed up the computation when running the plugin locally as the isClean-calculation is quite time consuming. Also it may help during development of the version-rules.

This solves one part of issue #736


bgalek:
LGTM, thx

PR #749, opened 2024-05-03
Add means to check if version is a snapshot programatically.

### LINKED ISSUE #678: How to check if current version is a snapshot programatically
How would one check if the current version is a snapshot programmatically?

Currently, I am using
```kotlin
val Project.isSnapshot: Boolean
    get() = version.toString().endsWith("-SNAPSHOT")
```
however, this feels sub-par.

There currently exists a class `VersionContext`, which contains the field `snapshot`, which is what I'd like to use, however there is no easy way to get this.
It would be nice if this was exposed publicly (Or some other way to check if the current version is a snapshot)

solonovamax:
Exposes cached `DecoratedVersion`, and introduce isSnapshot() to `DecoratedVersion`.

Fixes #678.

- Introduces a method, `getDecoratedVersion()` in the `VersionConfig` class, allowing you to easily retrieve the cached decorated version, instead of having to do `versionProvider().get()`, which looks less nice (imo) (and is especially nice in groovy/kotlin code where you'll do `decoratedVersion` instead of `versionProvider().get()`)
- Also introduces a method, `isSnapshot()` to `DecoratedVersion`, which allows you to check if the decorated version is a snapshot. This is passed the value of `VersionContact.isSnapshot()`. This is convenient for checking if a version is a snapshot or not when building the project.


solonovamax:
Do we need 100% code coverage, or no?

solonovamax:
@bgalek (you seem to be the main person merging prs)

bgalek:
hi! coverage is not a target - but having a tests is ;) Please add at least one test of this feature to document the behaviour

solonovamax:
it's mainly just exposing a couple of getters tbh, not anything very interesting, so idk how I'd go about testing that without said tests being effectively useless lol

also, could you take a peek at my other pr? because, I'm looking to get some feedback/opinions on it

solonovamax:
@bgalek yo, unsure if you saw this or not bc it was a weekend, so I'm gonna bump it lol

bgalek (approved):

PR #750, opened 2024-05-03
Add gradle task rule for marking the next version

### LINKED ISSUE #602: [question] can I create a shortcut gradle task to bump major\minor without args?
Hey, one of my little complaints and something that I think makes it a little bit more difficult for new learners is it needing to use the -Prelease=incrementMajor etc...

It doesn't help because gradle does not auto complete those, I looked around to see if that's possible but maybe it's just the way gradle is designed, that it cannot grab those info's for the IDE to auto complete on...

Did anyone else run into this concern and how could it be alleviated?

I was thinking it would just be easier for me to have a markNextReleaseMajor, markNextReleaseMinor and then it's just a simple easily discoverable gradle task. Gradle (kts ) doesn't seem to like tasks calling other tasks and passing command args, even though in this case it isn't a part of the execution, its intention is just a shortcut to a task we would normally run

Asking newcomers to simply run and remember these 2 tasks would be so much better than copying and pasting command line args and remembering those each time

Thoughts?

solonovamax:
Adds a gradle task rule for incrementing the next version, using the format `markNextVersion[Patch/Minor/Major/MinorIfNotRelease/Prerelease/Branch]`

Fixes #602

Downsides of this approach:
- Tasks are not present in the ide, so it's hard for users to know about them

Upsides of this approach:
- Does not pollute the task list with a bunch of tasks
- Tasks for incrementing the version are only created immediately before the task is executed
- No need to add new tasks if any new version incrementers are added

### DIFF
--- a/src/main/groovy/pl/allegro/tech/build/axion/release/BaseAxionTask.groovy
+++ b/src/main/groovy/pl/allegro/tech/build/axion/release/BaseAxionTask.groovy
@@ -19,7 +19,7 @@ abstract class BaseAxionTask extends DefaultTask {
     @Inject
     protected abstract ProviderFactory getProviders();
 
-    protected VersionResolutionContext resolutionContext() {
-        return VersionResolutionContext.create(versionConfig, layout.projectDirectory)
+    protected VersionResolutionContext resolutionContext(String nextVersion = null, String versionIncrementerName = null) {
+        return VersionResolutionContext.create(versionConfig, layout.projectDirectory, nextVersion, versionIncrementerName)
     }
 }
--- a/src/main/groovy/pl/allegro/tech/build/axion/release/MarkNextVersionTask.groovy
+++ b/src/main/groovy/pl/allegro/tech/build/axion/release/MarkNextVersionTask.groovy
@@ -1,15 +1,18 @@
 package pl.allegro.tech.build.axion.release
 
-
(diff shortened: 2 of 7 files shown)


solonovamax:
Marked as a draft, as I'm unsure if this is the best approach, instead of just creating several tasks.

in #602, I said:

> Hi, I've implemented this using task rules, however there is currently somewhat of a drawback:
> 
> it will not show the task in the ide.
> 
> with task rules, it allows you to specify the next marked version by doing
> 
> ```
> gradle run markNextVersion[Patch/Minor/Major/MinorIfNotRelease/Prerelease/Branch]
> ```
> 
> and then it will use the appropriate version incrementer. _however_, since these tasks technically don't actually exist until right before they're invoked, the ide has absolutely no knowledge of them. (basically you can tell gradle "if a task matches this pattern, then run this code right before attempting to execute the task", and the code that it runs creates the task with name of the task its about to run)
> 
> I'm debating whether or not I might just want to create a couple of dedicated tasks called
> 
>     * `markNextVersionPatch`
> 
>     * `markNextVersionMinor`
> 
>     * `markNextVersionMajor`
> 
>     * etc.
> 
> 
> that way, they'd actually show up inside of the gradle sidebar in intellij, for greater discoverability.
> 
> Opinions?

PR #760, opened 2024-05-16
improve ci_servers.md GitLab CI branch override

doofy:
Suggested branch override with CI_COMMIT_BRANCH is not possible on merge request pipelines as it would be empty (source: https://docs.gitlab.com/ee/ci/variables/predefined_variables.html). Override with CI_COMMIT_REF_SLUG to fix.


bgalek:
thx!

PR #772, opened 2024-07-05
set "VERSION_WITH_BRANCH" versionCreator as default

bgalek:
There is a common problem with collaboration on code using this plugin.

When multiple feature branches are being developed (diverged from the same HEAD), by default, they will use a SIMPLE version creator that does not decorate versions in any way. So, both feature branches will generate the same version (v0.0.1-SNAPSHOT), leading to overriding snapshots during publication. That is prone to errors since Maven allows overriding SNAPSHOT versions.
 
This PR changes the default versionCreator to VERSION_WITH_BRANCH, which will append the branch name to version (v0.0.1-my-branch-name-SNAPSHOT) and will fallback to SIMPLE for detached HEAD checkouts (old behavior) 

People who explicitly need the SIMPLE version will need to update their configurations.

Draft release

Screenshot 2024-07-05 at 10 11 03

### DIFF
--- a/docs/configuration/version.md
+++ b/docs/configuration/version.md
@@ -265,7 +265,7 @@ This rule uses additional parameter `initialPreReleaseIfNotOnPrerelease`
 ## Decorating
 
 Decorating phase happens only when version is read (and deserialized).
-During this phase version will be decorated with branch name.
+During this phase, the version will be decorated with a branch name (default behavior).
 `axion-release-plugin` supports adding predefined named version creators
 (so don't be afraid to post pull request if you have something useful!).
 Decoration phase is conducted by *version creators*,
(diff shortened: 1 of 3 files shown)


aleksandrserbin (approved):

docs/configuration/version.md
```diff
@@ -265,15 +265,15 @@ it's empty):
...
-During this phase version can be decorated with i.e. branch name.
-Default decorator does nothing. `axion-release-plugin` supports adding
-predefined named version creators (so don't be afraid to post pull
-request if you have something useful!). Decoration phase is conducted by
-*version creators*, you can configure it via `scmVersion.versionCreator`
+During this phase version will be decorated with branch name.
+`axion-release-plugin` supports adding predefined named version creators
+(so don't be afraid to post pull request if you have something useful!).
+Decoration phase is conducted by *version creators*,
+you can configure it via `scmVersion.versionCreator`
```
radoslaw-panuszewski, line 272: I would state more explicitly that default version creator is `versionWithBranch`

bgalek: check it now :)

radoslaw-panuszewski (approved):

PR #776, opened 2024-07-12
Read ref from GITHUB_HEAD_REF env variable if it's present

radoslaw-panuszewski:
When using `axion-release` together with [actions/checkout](https://github.com/actions/checkout), we needed to add the special configuration:
```
steps:
  - uses: actions/checkout@v4
    with:
      ref: ${{ github.head_ref }}
```
Otherwise, the action checks out repository in a detached-HEAD state and `axion-release` is not able to get the branch name.

Even though the branch ref is not checked out by default, Github provides the `GITHUB_HEAD_REF` which contains branch name that triggered the workflow. After this PR, `axion-release` will read this variable if available or fall back to git to preserve old behavior.

### DIFF
--- a/src/main/java/pl/allegro/tech/build/axion/release/infrastructure/git/GitRepository.java
+++ b/src/main/java/pl/allegro/tech/build/axion/release/infrastructure/git/GitRepository.java
@@ -324,14 +324,46 @@ public class GitRepository implements ScmRepository {
     }
 
     private String branchName() {
-        return branchNameFromEnvVariable().orElseGet(this::branchNameFromGit);
+        return branchNameFromGithubEnvVariable().orElseGet(this::branchNameFromGit);
     }
 
     /**
-     * @return head ref or source branch of the pull request in a workflow run
+     * @return branch name based on GitHub event that triggered the workflow
      */
-    private Optional<String> branchNameFromEnvVariable() {
-        return Optional.ofNullable(System.getenv("GITHUB_HEAD_REF"));
+    private Optional<String> branchNameFromGithubEnvVariable() {
+        return env("GITHUB_EVENT_NAME").flatMap(eventName -> {
+            switch (eventName) {
+                case "push":
+                    return branchNameForPushEvent();
+                case "pull_request":
+                case "pull_request_target":
+                    return branchNameForPullRequestEvent();
+                default:
+                    return Optional.empty();
+            }
+        });
+    }
+
+    private Optional<String> env(String name) {
+        return Optional.ofNullable(System.getenv(name));
+    }
+
+    /**
+     * Returns:
+     * <li>branch name if the branch ref triggered the workflow</li>
+     * <li>Optional.empty if tag ref triggered the workflow</li>
+     */
+    private Optional<String> branchNameForPushEvent() {
+        return env("GITHUB_REF_TYPE")
+            .filter(refType -> refType.equals("branch"))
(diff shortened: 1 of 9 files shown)


src/main/java/pl/allegro/tech/build/axion/release/infrastructure/git/GitRepository.java
```diff
@@ -333,37 +304,54 @@ private void assertPathExists(String path) {
...
-            String branchName = branchName();
-            return new ScmPosition(revision, branchName, isClean);
+            return revision;
         } catch (IOException e) {
             throw new ScmException(e);
         }
     }
 
+    private String branchName() {
+        return branchNameFromEnvVariable().orElseGet(this::branchNameFromGit);
```
bgalek, line 327: @radoslaw-panuszewski could you add docs entry about this happening?

bgalek (approved):

src/main/java/pl/allegro/tech/build/axion/release/infrastructure/git/GitRepository.java
```diff
@@ -333,37 +304,54 @@ private void assertPathExists(String path) {
...
+    }
+
+    /**
+     * @return head ref or source branch of the pull request in a workflow run
+     */
+    private Optional<String> branchNameFromEnvVariable() {
+        return Optional.ofNullable(System.getenv("GITHUB_HEAD_REF"));
```
bgalek, line 334: We need to handle multiple trigger types here (based on GITHUB_EVENT_NAME):
`push` -> GITHUB_REF_NAME
`pull-request`, `pull_request_target` -> GITHUB_HEAD_REF

We should only apply this logic where GITHUB_REF_TYPE is `branch` probably too.

Sources: 
https://docs.github.com/en/actions/learn-github-actions/variables
https://docs.github.com/en/actions/learn-github-actions/variables

bgalek (changes requested):
We need to handle multiple trigger types here (based on GITHUB_EVENT_NAME):
push -> GITHUB_REF_NAME
pull-request, pull_request_target -> GITHUB_HEAD_REF

We should only apply this logic where GITHUB_REF_TYPE is branch probably too.

Sources:
https://docs.github.com/en/actions/learn-github-actions/variables
https://docs.github.com/en/actions/learn-github-actions/variables

bgalek:
@radoslaw-panuszewski nice ;) 
It looks like workdlow_dispatch|call|run have GITHUB_HEAD_REF set and it cannot be committed since the UI requires you to select branch

bgalek:
last thing: please update docs :)

src/main/java/pl/allegro/tech/build/axion/release/infrastructure/git/GitRepository.java
radoslaw-panuszewski: Done

radoslaw-panuszewski:
We don't need to `GITHUB_REF_NAME` env variable for `push` events, because in case for those events, GitHub checks out either branch or tag that was pushed

bgalek (approved):

bgalek:
thx, that looks very nice!

Zordid:
I never used to do this ref thing in my GitHub actions... it just worked. Do I now have to change anything coming from 1.18.0?

bgalek:
@Zordid nope, it should be 100% backward compatibile

PR #793, opened 2024-08-16
Implement support for fallback tag prefixes

radoslaw-panuszewski:
### Usage

```kotlin
scmVersion {
    tags {
        prefix = "v"
        fallbackPrefixes = listOf("${rootProject.name}-")
    }
}
```

### How it works

We can think of `fallbackPrefixes` as "read-only prefixes". They are used as fallbacks when deserializing version from tags, but only the "main" prefix will be serialized. Even during deserialization, the "main" prefix takes precedence over the fallback prefixes.

### DIFF
--- a/docs/configuration/version.md
+++ b/docs/configuration/version.md
@@ -33,6 +33,17 @@ calculating current version. Prefix can be set using
 
 Default prefix is `v`.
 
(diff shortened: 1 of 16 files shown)


bgalek (approved):

bgalek:
LGTM, hopefully it will ease the migration!

PR #830, opened 2024-10-04
Ignore non-semver tags

radoslaw-panuszewski:
I'm not sure if it's worth to hide this behavior under a config property - the only difference is that the plugin won't fail in some cases and just ignore the non-semver tags

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,6 +4,7 @@ root = true
 [*]
 indent_style = space
 indent_size = 4
+ij_continuation_indent_size = 4
 end_of_line = lf
 charset = utf-8
 trim_trailing_whitespace = true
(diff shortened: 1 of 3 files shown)


bgalek (approved):

vladimirfx:
In which cases this should work?

Example:

Prefix: `mfc-backend`
Latest tag: `mfc-backend-5.21.1`
Old tag (6 year old): mfc-backend-core-2.44.1

Error:

```log
❯ ./gradlew cV

FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring root project 'mfc-backend'.
> Failed to parse version: mfc-backend-core-2.44.1 that matched configured prefix: mfc-backend. There can be no tags that match the prefix but contain non-SemVer string after the prefix. Detailed message: Unexpected character 'LETTER(c)' at position '0', expecting '[DIGIT]'

```

I can't use that prefix because 6 years ago the module name was longer. It will be cool if incorrect tags are just ignored.

Issue #826, opened 2024-10-01
[Feature / suggestion] Print warning when no prior version is detected


ddeath:
> I saw it couple times in our project, sometimes the plugin probably does not detect any tag or something and it will create default initial version of `0.1.0` even though there are 10 prior tags. This is happening on Gitlab CI pipeline.

I found the issue. Basically default behavior of Gitlab is that it will fetch (not clone) repository and on top of that only top 20 commits. So if you have maybe monorepo, or for some any other reason you didnt release in last 20 commits, axion will not find any tag tight to those commits and will assume this is the first release and will use 0.1.0 version.
Above situation will usually result in error:
```
remote status: REJECTED_NONFASTFORWARD
remote message: 
> Task :release FAILED
```

Because axion will try to push tag 0.1.0 but it already exists.

This was kind of hard to debug therefore I would suggest to print `Didn't detect any previous version in available git history, treating it as new release` which will get more explicit hint to the developer.

Ideally I would add some debug / verbose mode where plugin would also tags it sees and then all matching tags etc...

tomasz-pankowski-allegro:
spojrzę na to zadanie w ramach hacktoberfest

oallauddin:
Please update the GitLab CI example in the documentation.  I was running into this issue also.  
https://axion-release-plugin.readthedocs.io/en/latest/configuration/ci_servers/#gitlab-ci

Adding GIT_STRATEGY and GIT_DEPTH resolved the problem for me.  
https://docs.gitlab.com/ee/ci/runners/configure_runners.html#git-strategy
https://docs.gitlab.com/ee/ci/runners/configure_runners.html#shallow-cloning
```
  variables:
    GIT_STRATEGY: clone 
    GIT_DEPTH: 0
```

PR #844, opened 2024-10-16
Add verbose information to version generation logic

tomasz-pankowski-allegro:
Fix for https://github.com/allegro/axion-release-plugin/issues/826

Adding information about missing tags

### DIFF
--- a/docs/configuration/version.md
+++ b/docs/configuration/version.md
@@ -427,6 +427,6 @@ property:
     }
 
 ## Troubleshooting
-In case of problems use can try to use verbose mode:
+In case of problems you can try to use verbose mode:
 
     #./gradlew release -Prelease.verbose
--- a/docs/configuration/version.md
+++ b/docs/configuration/version.md
@@ -425,8 +425,3 @@ property:
     scmVersion {
         sanitizeVersion.set(false)
     }
-
-## Troubleshooting
-In case of problems you can try to use verbose mode:
-
-    #./gradlew release -Prelease.verbose
--- a/docs/configuration/version.md
+++ b/docs/configuration/version.md
@@ -425,3 +425,8 @@ property:
     scmVersion {
         sanitizeVersion.set(false)
     }
+
+## Troubleshooting
+In case of problems use can try to use verbose mode:
+
+    #./gradlew release -Prelease.verbose
--- a/src/main/java/pl/allegro/tech/build/axion/release/domain/VersionSorter.java
+++ b/src/main/java/pl/allegro/tech/build/axion/release/domain/VersionSorter.java
@@ -80,14 +80,7 @@ class VersionSorter {
             logger.quiet("No tags were found in git history");
         }
 
-        List<Version> versionList = new ArrayList<>(versions);
-        Collections.sort(versionList, Collections.reverseOrder());
-
-        Version version = (versionList.isEmpty() ? versionFactory.initialVersion() : versionList.get(0));
-        if (version == null) {
-            version = versionFactory.initialVersion();
-        }
-
+        Version version = getLatestVersion(versionFactory, versions);
         logger.debug("Version: {}, was selected from versions: {}", version, versions);
 
         TagsOnCommit versionCommit = versionToCommit.get(version);
@@ -98,6 +91,13 @@ class VersionSorter {
             versions.isEmpty(),
             (versionCommit == null ? null : versionCommit.getCommitId())
         );
+    }
+
+    private static Version getLatestVersion(VersionFactory versionFactory, Set<Version> versions) {
+        return versions.stream()
+            .filter(Objects::nonNull) // filter out nulls
+            .max(Comparator.naturalOrder()) // find latest version
+            .orElseGet(versionFactory::initialVersion); // if no versions found, return initial version
     }
 
     static class Result {
--- a/src/main/java/pl/allegro/tech/build/axion/release/domain/VersionSorter.java
+++ b/src/main/java/pl/allegro/tech/build/axion/release/domain/VersionSorter.java
@@ -80,7 +80,14 @@ class VersionSorter {
             logger.quiet("No tags were found in git history");
         }
 
-        Version version = findBestVersion(versionFactory, versions);
+        List<Version> versionList = new ArrayList<>(versions);
+        Collections.sort(versionList, Collections.reverseOrder());
+
+        Version version = (versionList.isEmpty() ? versionFactory.initialVersion() : versionList.get(0));
+        if (version == null) {
+            version = versionFactory.initialVersion();
+        }
+
         logger.debug("Version: {}, was selected from versions: {}", version, versions);
 
         TagsOnCommit versionCommit = versionToCommit.get(version);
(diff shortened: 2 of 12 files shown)


src/main/java/pl/allegro/tech/build/axion/release/domain/VersionSorter.java
```diff
@@ -21,6 +23,12 @@
  * * any highest (stable or alpha)
  */
 class VersionSorter {
+    private static final Logger logger = Logging.getLogger(VersionSorter.class);
+    private final boolean verbose;
+
+    public VersionSorter(boolean verbose) {
```
tomasz-pankowski-allegro, line 29: isn't it too ugly?

src/main/java/pl/allegro/tech/build/axion/release/domain/VersionSorter.java
```diff
@@ -47,10 +57,16 @@ Result pickTaggedCommit(
             for (String tag : tags) {
                 boolean isNextVersion = nextVersionTagPattern.matcher(tag).matches();
                 if (isNextVersion && (ignoreNextVersionTags || ignoreNextVersionOnHead)) {
+                    if (verbose) {
```
ddeath, line 60: Instead of using `IFs` we should just do `logger.debug()`
https://docs.gradle.org/current/userguide/logging.html

src/main/java/pl/allegro/tech/build/axion/release/domain/VersionSorter.java
```diff
@@ -47,10 +57,16 @@ Result pickTaggedCommit(
...
+                        logger.quiet("Ignoring tag: {}, because it's a next version tag and it's not forced", tag);
```
tomasz-pankowski-allegro, line 61: I'm not sure which login level would be most appropriate

src/main/java/pl/allegro/tech/build/axion/release/domain/VersionSorter.java
```diff
@@ -32,9 +40,11 @@ Result pickTaggedCommit(
         Set<Version> versions = new LinkedHashSet<>();
         LinkedHashMap<Version, Boolean> isVersionNextVersion = new LinkedHashMap<>();
         LinkedHashMap<Version, TagsOnCommit> versionToCommit = new LinkedHashMap<>();
+        boolean tagsFound = false;
 
         for (TagsOnCommit tagsEntry : taggedCommits.getCommits()) {
             List<String> tags = tagsEntry.getTags();
+            tagsFound = tagsFound || !tags.isEmpty();
```
ddeath, line 47: When this will happen? Since we are iterating through `taggedCommits` I would assume that each commit has at least 1 tag otherwise it would NOT be `tagged` 🤔

docs/configuration/version.md
```diff
@@ -425,3 +425,8 @@ property:
     scmVersion {
         sanitizeVersion.set(false)
     }
+
+## Troubleshooting
+In case of problems you can try to use verbose mode:
+
+    #./gradlew release -Prelease.verbose
```
ddeath, line 432: See comment bellow. We should probably do just `./gradlew [command] -d`, no need for passing custom flag https://docs.gradle.org/current/userguide/logging.html

src/main/java/pl/allegro/tech/build/axion/release/domain/VersionSorter.java
tomasz-pankowski-allegro: yes, I was thinking about it, but either we'll have to enable debug everywhere, which will probably clutter the logs a lot, but maybe that's ok? or we'll have to indicate for which class we want to change the logging level, which is also probably doable, but maybe less user friendly? What do you think?

src/main/java/pl/allegro/tech/build/axion/release/domain/VersionSorter.java
ddeath: but what `verbose` actually means? Some random google definition:

> using or expressed in more words than are needed.

So I would not be worried about that. Also currently there seems to be only `4` more debug logs:
https://github.com/search?q=repo%3Aallegro%2Faxion-release-plugin%20logger.debug&type=code

src/main/java/pl/allegro/tech/build/axion/release/domain/VersionSorter.java
tomasz-pankowski-allegro: I think that currently in the code there is actually no way for this class to have an empty tag collection, but there is also nothing in this class that would prevent the creation of such an instance. This class even has a special method for such an occasion ;) TagsOnCommit.empty() which creates an instance with an empty list of tags.
This code, I hope, will work correctly in both cases:
* if taggedCommits.getCommits() is empty tagsFound==false
* if the list of tags in taggedCommits.getCommits() is empty then tagsFound==false
The advantage is that we don't have to wonder if someone really thought it was a great idea to pass an empty list to the TagsOnCommit constructor.
You can also set a contract for TagsOnCommit, the list of tags cannot be and then we throw an exception otherwise. Then you can easily simplify this code as you say.
But I'm not insisting, if you feel that this is an unnecessary complication then I'll let you convince me.

src/main/java/pl/allegro/tech/build/axion/release/domain/VersionSorter.java
tomasz-pankowski-allegro: I'm more afraid of debug logs from the entire gradel tool + everything else we call at the same time. But I haven't checked what the result would look like now with debug turned on, I can check it so we don't guess

tomasz-pankowski-allegro: I did fast test number o lines:
./gradlew release -Prelease.dryRun -d > debug.log
    6877 debug.log
    
 ./gradlew release -Prelease.dryRun > standart.log
      27 standart.log

tomasz-pankowski-allegro: but maybe it's still ok, maybe it's better than complicating the code, and if someone needs additional information, they will somehow dig through these thousands of lines of logs?

src/main/java/pl/allegro/tech/build/axion/release/domain/VersionSorter.java
```diff
@@ -68,14 +84,13 @@ Result pickTaggedCommit(
 
             }
         }
+        if (!tagsFound) {
+            logger.quiet("No tags were found in git history");
```
ddeath, line 88: imho this should be warning

src/main/java/pl/allegro/tech/build/axion/release/domain/VersionSorter.java
tomasz-pankowski-allegro: I did a quick research and checked in a few other plugins if someone uses the verbose flag or rather the login level. From what I see, however, the login level is used, so I will choose this option. It will be much much simpler

src/main/java/pl/allegro/tech/build/axion/release/domain/VersionSorter.java
radoslaw-panuszewski: No, the `lifecycle` log level is the default log level [(docs](https://docs.gradle.org/current/userguide/logging.html)). If you log on `quiet`, it will be visible even when running with `./gradlew -q`. It's not a good practice to log everything on `quiet` since users that explicitly want to hide the logs via `-q` switch will see them anyway.

bgalek: my mistake, I forgot about `lifecycle`

tomasz-pankowski-allegro:
I have the impression that there is no consensus on how this should be done, so I think it would be better to close this PR

PR #854, opened 2024-10-29
Changed the compilation jdk from 11 to 17

Lakhwinderr:
Hi, I am beginner to this tech stack. As per my knowledge, I changed the JDK version to 17. Do let me know other changes to make.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -23,7 +23,7 @@ java {
     withSourcesJar()
     withJavadocJar()
     toolchain {
-        languageVersion.set(JavaLanguageVersion.of(11))
+        languageVersion.set(JavaLanguageVersion.of(17))
     }
 }
 


solonovamax:
What is the benefit of this?

Currently, [gradle supports jdk 8-23](https://docs.gradle.org/current/userguide/compatibility.html), and support for jdks 8-16 will not be dropped until gradle 9.0 is released.
You are able to run the plugin just fine when using a jdk ≥11 for gradle, currently.

Zordid:
Also don't see the point in such a PR...

bgalek:
Sooner o later people need to migrate to later JDK's ;)

Zordid:
I meant in such a tiny PR with changing a number. No other changes, no tests and also stating he's a beginner. I would not trust in that he knows the consequences of changing this for a product used by many. He did not even adjust the README.md to reflect what he did. So, I stand with my comment. What's the point in doing such a PR? Did he think nobody else was able to change that number before?

bgalek:
@Zordid it was hacktoberfest contribution, i should have made a better issue description, it's on me

solonovamax:
> Someday people need to migrate to later JDK's anyway ;)
> I'm going to switch to JDK21 for sure, but I'll leave compilation target to JDK11 for now

you should keep targeting jdk 8 until gradle 9 is released, at which point the minimum gradle jdk will be bumped from 8 to 17 (see: https://gradle.org/whats-new/gradle-9/)

sometimes it is still necessary to run gradle with a java 8 jdk, so support for it should not be dropped unless absolutely necessary.

bgalek:
Another reason: https://github.com/allegro/axion-release-plugin/pull/911 :)

Issue #500, opened 2022-06-22
KotlinDSL preRelease fileUpdate cannot be cast to groovy.lang.Closure


JayBur:
Hi, 
Inspired to using Kotlin DSL and the recent additions to the plugin I started using the new Kotlin DSL methods. They seem to work except for fileUpdate. I considered writing this in #449 , however since it's released I decided this might be better reported on its own.

With v1.13.14 I get a ClassCastException `Build_gradle$1$1$3$1$1 cannot be cast to groovy.lang.Closure` for the following fileUpdate:

```kotlin
scmVersion {
    hooks {
         // [..] logging and custom context calls for addCommitPattern()

        preRelease {
            fileUpdate {
                file("src/main/resources/specs/some_api.yaml")
                file("src/main/resources/specs/some_other_api.yaml")
                pattern = { previousVersion, _ -> "version: '$previousVersion'" }
                replacement = { currentVersion, _ -> "version: '$currentVersion'" }
            }
        }

        pre("commit")
    }
}
```
Full stacktrace:
```
* Exception is:                                                                                                                                   
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':release'.                                                                
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.lambda$executeIfValid$1(ExecuteActionsTaskExecuter.java:147)        
        at org.gradle.internal.Try$Failure.ifSuccessfulOrElse(Try.java:282)                                                                       
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeIfValid(ExecuteActionsTaskExecuter.java:145)                 
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:133)                        
        at org.gradle.api.internal.tasks.execution.CleanupStaleOutputsExecuter.execute(CleanupStaleOutputsExecuter.java:77)                       
        at org.gradle.api.internal.tasks.execution.FinalizePropertiesTaskExecuter.execute(FinalizePropertiesTaskExecuter.java:46)                 
        at org.gradle.api.internal.tasks.execution.ResolveTaskExecutionModeExecuter.execute(ResolveTaskExecutionModeExecuter.java:51)             
        at org.gradle.api.internal.tasks.execution.SkipTaskWithNoActionsExecuter.execute(SkipTaskWithNoActionsExecuter.java:57)                   
[...]

PR #891, opened 2025-02-16
Fix Kotlin DSL not being compatible with internal hooks implementation

danilaslynko:
Issue #500


bgalek (approved):

bgalek:
LGTM, I'll need to check what happend with staging repository setting on CI

PR #904, opened 2025-03-12
Defaulting to root's build directory when used from `includedBuild`s

gmazzo:
This contemplates the case when the plugin is applied from an included build, by defaulting to root's build directory.

This is NO-OP for regular usages, but for those cases, it will remove the need of doing:
```kotlin
scmVersion.repository.directory.set("../")
```

### DIFF
--- a/src/main/groovy/pl/allegro/tech/build/axion/release/ReleasePlugin.groovy
+++ b/src/main/groovy/pl/allegro/tech/build/axion/release/ReleasePlugin.groovy
@@ -2,12 +2,15 @@ package pl.allegro.tech.build.axion.release
 
 import org.gradle.api.Plugin
 import org.gradle.api.Project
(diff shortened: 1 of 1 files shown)


bgalek:
just to make sure, did you test it with older gradle versions? (I need to setup matrix tests some day)

gmazzo:
> this is really good idea, thx!
> 
> just to make sure, did you test it with older gradle versions? (I need to setup matrix tests some day)

I think this is a very old API, so yes. But what minimum Gradle version are you supporting? I may try to add a test for min and current.

bgalek:
yup, should be fine, thanks!

bgalek (approved):

bgalek:
@gmazzo released ;)
https://github.com/allegro/axion-release-plugin/releases/tag/v1.18.18

Issue #453, opened 2022-05-03
When releasing a new tag, include a message with the commit info


roschwartz79:
When pushing a new tag, it's nice to see a short description added along with the tag. Even if it is just relaying the commit message that would be awesome. I'd like to note I am using the plugin in a CI environment, so maybe there's some extra configuration that could be added so it could be the users choice?

What the tags look like: 
<img width="346" alt="image" src="https://user-images.githubusercontent.com/44881235/166557789-67ab32b3-ee4f-4626-873d-a952c356ed20.png">

What I think would be helpful (roughly)
<img width="172" alt="image" src="https://user-images.githubusercontent.com/44881235/166558353-49a0ce65-2248-4ca1-ad03-46bae95e6d0b.png">

bgalek:
Hi, axion-release-plugin only creates git tags.
If you want a nice message and release notes you need to use github releases (https://github.com/allegro/axion-release-plugin/releases) ;)

roschwartz79:
> Hi, axion-release-plugin only creates git tags. If you want a nice message and release notes you need to use github releases (https://github.com/allegro/axion-release-plugin/releases) ;)

Sure, I know that's how you go about creating a nice release with all the notes, but can't you add a message with a tag too?

roschwartz79:
@bgalek I think I wasn't clear enough- I am looking to add annotations on my tags. However in the tests I noticed that annotations are always set to false. So maybe this isn't possible. But I would like to add annotations if that makes more sense!

bgalek:
Hm... We could provide an optional method/parameters to customize tag annotations. It would be backward compatible, care for a PR?

roschwartz79:
That's what I was thinking. I've just recently gotten into developing gradle plugins so I'll give it a shot and see what I come up with!

sreich:
@roschwartz79  hey there, have you been able to make some progress on this? this is a feature i would like to have

ebAtUelzener:
Hello there, i've taken the liberty to take a shot at implementing a feature to allow setting the tag message.

Please feel free to leave feedback :)

PR #1006, opened 2026-02-09
Feature: tag message from config (#453)

ebAtUelzener:
This PR implements a way to set the tag message via a callback.

It does not change default behaviour of the plugin.

The explicit suggestion from the original issue (relaying the commit message) is *not* intended to be implemented by this, however it is possible to provide extra information from CI using e.g. Gradle properties/ environment variables and reading these to set the message.

Minimal example reading an environment variable:

```kotlin
scmVersion {
  tag {
    messageCreator { position, version, previousVersion ->
      val envVar = System.getenv("SOME_VAR")
      "This is a message on version $version (previous: $previousVersion) {commit: ${position.shortRevision}} [$envVar]"
    }
  }
}
```

This would allow users to use arbitrary logic to set a tag message, without changing too much of the plugin's internals.


bgalek (approved):
thx! lgtm

PR #1074, opened 2026-08-28
Skip the working tree scan when uncommitted changes are ignored

FireHook:
### Problem

`VersionResolver` builds `ScmState` with an eager call:

```java
new ScmState(..., repository.checkUncommittedChanges())
```

but `VersionFactory` only ever uses it behind a guard:

```java
boolean hasUncommittedChanges =
    !versionProperties.isIgnoreUncommittedChanges() && scmState.hasUncommittedChanges();
```

`ignoreUncommittedChanges` defaults to `true`, so on a default build the working tree scan is computed and immediately thrown away. `&&` would short-circuit it if the value were lazy.

That scan is `jgitRepository.status().call()`, which stats the whole working tree. On a large repository it is one of the most expensive things axion does during configuration.

### Change

`ScmState` now holds a memoized `BooleanSupplier` instead of a `boolean`, and `VersionResolver` passes `repository::checkUncommittedChanges`. Nothing else changes: the value is still computed at most once, and still computed eagerly whenever `ignoreUncommittedChanges = false`.

`toString()` deliberately does not force the check, so logging the state stays cheap.

### Compatibility

The `boolean` constructor is kept and `@Deprecated`, delegating to the supplier one, so existing source and binary callers keep working.

One visible side effect: with `ignoreUncommittedChanges = true` (the default), `ScmRepository.checkUncommittedChanges()` is no longer called during version resolution at all. In `--dry-run`, `NoOpRepository`'s `"uncommitted changes: ..."` log line therefore no longer appears in that case. Happy to keep the call if you consider that log part of the contract, but it seemed like the point of the change.

### Measurements

Isolated benchmark against a 1 GB pack / 169,177 commit / 745 tag repository, resolving `currentVersion` with a warm daemon, alternating runs:

| | before | after |
|---|---|---|
| configuration time | 30.6s | **24.5s** |
| working tree scans per resolution | 14 | **10** |

The remaining 10 come from `currentPosition()` computing `isClean` eagerly; that is a separate discussion since `isClean` is a user-visible `@Input`.

### Tests

`VersionResolverTest` asserts the difference rather than an absolute count, because `currentPosition()` also calls the method:

```groovy
checking.uncommittedChangesChecks == ignoring.uncommittedChangesChecks + 1
```

### DIFF
--- a/src/main/java/pl/allegro/tech/build/axion/release/domain/ScmState.java
+++ b/src/main/java/pl/allegro/tech/build/axion/release/domain/ScmState.java
@@ -1,17 +1,30 @@
 package pl.allegro.tech.build.axion.release.domain;
 
+import java.util.function.BooleanSupplier;
+
 public class ScmState {
 
     private final boolean onReleaseTag;
     private final boolean onNextVersionTag;
     private final boolean noReleaseTagsFound;
-    private final boolean hasUncommittedChanges;
+    private final BooleanSupplier uncommittedChanges;
 
-    public ScmState(boolean onReleaseTag, boolean onNextVersionTag, boolean noReleaseTagsFound, boolean hasUncommittedChanges) {
+    private Boolean memoizedUncommittedChanges;
+
+    public ScmState(boolean onReleaseTag, boolean onNextVersionTag, boolean noReleaseTagsFound, BooleanSupplier uncommittedChanges) {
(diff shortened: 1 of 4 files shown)


bgalek:
@FireHook thx for the PR, I find it reasonable to do it like that, LGTM

bgalek (approved):

Issue #182, opened 2017-05-29
Negative impact on configuration phase performance


tmszdmsk:
We use `axion-release-plugin` in our git monorepo which consists of over 60 deployables. Profiling our build process revealed that `axion-release-plugin` is responsible for most of execution time in configuration phase. With `axion-release-plugin` enabled it takes about 700ms, while w/o it is ~15ms. Of course if we multiply it by 60 it turns out that plugin adds ~45seconds to total build time of the whole repository. 

1) Is there any way to speed it up?
2) If it is ran multiple times in the context of one repo(as in our case) - can any optimisation be made to make it effectively run once (and cache some data between executions)?

adamdubiel:
I never ran any performance tests against axion-release, since we use single repos, so I think there is plenty of room for improvement. Could you generate sample monorepo project for test purposes?

tmszdmsk:
@adamdubiel 
So... I think monorepo itself isn't an issue. I believe that JGit has some performance issues with big repositories (our has ~20k commits in master). I did an example repository based on axion-release-plugin and added a couple of commits on top of it (20k).

My totally unprofessional benchmark shows that `./gradlew build --dry-run --profile` configuration phase finishes in:
* ~0.05s for original `axion-release-plugin` repo
* ~0.2s for https://github.com/tmszdmsk/axion-long-history-repo
I ran it multiple times for each repo and picked lowest values I could get.

adamdubiel:
Interesting.. Are you using [tag with highest version](http://axion-release-plugin.readthedocs.io/en/latest/configuration/version.html#tag-with-highest-version) mode or the default?

tmszdmsk:
this is the whole config we use (so the answer is "default")
```
scmVersion{
    tag {
        prefix = 'appname'
    }
}
```

tmszdmsk:
I've dug into the code a little bit and I have an impression that retrieving `scmVersion.version` (which is all you need in configuration phase) does few more things "by the way" that enormously affect the current version resolution time. IMHO there should by some fast lane to retrieve that property as fast as possible. 

I can work on that issue in my spare time but I think I need some guidance at the beginning. What do you think?

adamdubiel:
I started writing JMH test to profile `GitRepository` class. Simple test shows that it takes ~100ms to  fetch stuff from Git in empty repository. I didn't yet test it in huge (20000k commits) repo.

What are the "by the way" things that you found?

tmszdmsk:
[...]

PR #1076, opened 2026-08-31
Optional native git backend for read operations on large repositories

### LINKED ISSUE #1075: Optional native git backend for read operations on large repositories
### Context

Follow-up to #736 and #182, both of which reported that version resolution dominates the configuration phase on large repositories.

#736 proposed three fixes. Only the first landed:

- ✅ an override flag for `isClean` → `overriddenIsClean` (#543)
- ❌ *"alternatively make the isClean-property to be lazily evaluated, so it's only computed if it is needed"*
- ❌ *"retrieving the tags takes some time because it iterates the whole log ... maybe it is enough to partial-compute the data, only for the relevant tags"*

I hit the same wall and implemented the two remaining suggestions, plus an optional native-git read path. #1074 covers the lazy `isClean` part. This issue is about the rest, since it adds public DSL surface and I would rather agree on the approach before opening a ~1000 line PR.

### Measurements

Repository: 1 GB pack, 169,177 commits, 745 tags, ~200 Gradle modules.

Measuring inside the real build was useless — the configuration phase swings between 50s and 104s regardless of backend, because the Android plugins dominate. So I benchmarked in an isolated single-module project pointing at the same `.git`, warm daemon, alternating runs. That is reproducible to within 30 ms.

| | JGit | native git |
|---|---|---|
| `main` today | 30.6s | — |
| with #1074 | 24.5s | — |
| with #1074 + native backend | — | **19.0s** |

Cost of the underlying commands on that repository:

```
git rev-list HEAD              0.657s
git for-each-ref refs/tags     0.038s
git status --porcelain         1.599s
```

Note that tags are already packed here, so the `git gc` / `packed-refs` workaround from #736 does not apply — the walk itself is the cost.

### Proposal: hybrid backend, opt-in

```groovy
scmVersion {
    repository {
        backend.set("nativeGit")   // default: "jgit"
    }
}
```

plus `-Prelease.scmBackend=nativeGit`. Default stays `jgit`, so no behaviour change unless opted in.

`ScmRepositoryFactory` picks the implementation; a new `NativeGitRepository` implements the read half of `ScmRepository` and **delegates every write to the existing JGit `GitRepository`**.

### Why hybrid rather than a full native rewrite

A full rewrite would also have to reimplement push/fetch/commit and the remote authentication that `TransportConfigCallback` / `ScmIdentity` does today (ssh keys, in-memory PGP, token auth). Replacing that with ambient credential and ssh config would be a real, security-sensitive behaviour change. Since all of the measured cost is in reads, delegating writes keeps release tagging and pushing bit-for-bit unchanged.

### Implementation notes

- **Configuration cache.** Reads go through `ProviderFactory.exec`, not `ProcessBuilder` — raw process execution during configuration is a hard configuration-cache failure ("external process started ...").
- **Bounded tag walk.** This is #736's "partial-compute" suggestion. `providers.exec` cannot stream, so the walk reads a bounded chunk (`rev-list --max-count=2000`) and only falls back to the full history if that chunk yields no match *and* more history exists. Exact, not approximate. On this repository the nearest tag is 103 commits from HEAD, and the bounded call costs 0.042s versus 0.660s.
- **Semantics preserved.** Nearest reachable tagged commit, annotated tags peeled via `%(*objectname)`, `for-each-ref` ref-name ordering matching `tagList()`, `HEAD` when detached, `overriddenBranchName` / `GITHUB_HEAD_REF` / `overriddenIsClean` / `releaseBranchNames` all honoured, `:(exclude)` pathspecs for `excludeSubFolders`.
- **Hermetic.** `GIT_CONFIG_GLOBAL` / `GIT_CONFIG_SYSTEM` are neutralised when `ignoreGlobalGitConfig` is set, mirroring `SystemReaderWithoutSystemConfig`. `GIT_OPTIONAL_LOCKS=0` avoids taking `index.lock` (measured: no cost).
- **Requires** `git` on `PATH` when enabled.

### Alternatives considered

- **Caching `GitRepository` results** — discussed at length in #182. It would help both backends and cut more than this does: resolving one version currently issues 70 git operations, including 14 working-tree scans and 8 tag walks. But it changes staleness semantics after `release` creates a tag, and #182 notes the plugin is applied per-module with different tag prefixes, which complicates the cache key. Happy to explore this instead if you prefer.
- **`overriddenIsClean`** already avoids the status cost, but it is a correctness trade-off users must opt into per-project; #1074 makes it unnecessary in the default case.

### Honest caveats

- This only pays off on large repositories. Resolving a version spawns roughly a dozen `git` processes, so on small repositories the native backend is a wash or marginally slower. That is why it is opt-in.
- The win only materialises on configuration-cache **misses**. On a cache hit axion does not run at all.
- The remaining hotspot after both changes is `currentPosition()` recomputing `isClean` (10 of the 14 scans). Fixing that means making `ScmPosition.isClean` lazy, which touches a user-visible `@Input` — deliberately out of scope here.

### Status

Implemented and green: `NativeGitRepositoryTest` mirrors the read-path scenarios of `GitRepositoryTest` and additionally asserts output parity with the JGit backend on the same fixture repository, plus integration tests for the DSL flag, the Gradle property and an unknown backend value.

Would you accept a PR along these lines? I am happy to change the DSL naming (`backend` vs something else), make it a Gradle-property-only escape hatch rather than DSL, or drop it in favour of the caching approach from #182.

FireHook:
Closes #1075.

Follow-up to #1074, and to the two suggestions in #736 that never landed:

> *"retrieving the tags takes some time because it iterates the whole log ... maybe it is enough to partial-compute the data, only for the relevant tags"*

### What this does

Adds an **opt-in** read path backed by the native `git` executable. Default stays JGit, so there is no behaviour change unless it is switched on.

```groovy
scmVersion {
    repository {
        backend.set("nativeGit")   // default: "jgit"
    }
}
```

or `-Prelease.scmBackend=nativeGit` (works from `gradle.properties` too).

`ScmRepositoryFactory` picks the implementation. `NativeGitRepository` implements the read half of `ScmRepository` and **delegates every write to the existing JGit `GitRepository`** — `fetchTags`, `tag`, `dropTag`, `push`, `commit`, `attachRemote` are untouched.

### Why hybrid rather than a full native rewrite

A full rewrite would also have to reimplement the remote authentication `TransportConfigCallback` / `ScmIdentity` performs today (ssh keys, in-memory PGP, token auth). Relying on ambient credential and ssh config instead would be a real, security-sensitive behaviour change. All of the measured cost is in reads, so delegating writes keeps release tagging and pushing bit-for-bit identical.

### Measurements

1 GB pack, 169,177 commits, 745 tags, isolated single-module project against the same `.git`, warm daemon, alternating runs (reproducible to ~30 ms):

| | JGit | native git |
|---|---|---|
| before #1074 | 30.6s | — |
| with #1074 (current `main`) | 24.5s | — |
| this PR | — | **19.0s** |

### Notes for review

- **Configuration cache.** Reads go through `ProviderFactory.exec`, not `ProcessBuilder` — raw process execution during configuration is a hard CC failure (`external process started ...`). This is why `ScmProperties` carries a `ProviderFactory`.
- **Bounded tag walk.** `providers.exec` cannot stream, so the walk reads a bounded chunk (`rev-list --max-count=2000`) and falls back to the full history only if that chunk yields no match *and* more history exists. Exact, not approximate. On the measured repo the nearest tag is 103 commits from HEAD; the bounded call costs 0.042s vs 0.660s. A package-private constructor takes the chunk size so both paths are tested without a huge fixture.
- **Semantics preserved.** Nearest reachable tagged commit; annotated tags peeled via `%(*objectname)`; `for-each-ref` ref-name ordering matches `tagList()`; `HEAD` when detached; `overriddenBranchName`, `GITHUB_HEAD_REF`, `overriddenIsClean`, `releaseBranchNames` all honoured; `:(exclude)` pathspecs for `excludeSubFolders`.
- **Hermetic.** `GIT_CONFIG_GLOBAL` / `GIT_CONFIG_SYSTEM` neutralised when `ignoreGlobalGitConfig` is set, mirroring `SystemReaderWithoutSystemConfig`. `GIT_OPTIONAL_LOCKS=0` avoids taking `index.lock` (measured: no cost).
- **Requires** `git` on `PATH` when enabled.

### Tests

`NativeGitRepositoryTest` mirrors the read-path scenarios of `GitRepositoryTest` and additionally asserts **output parity with the JGit backend** on the same fixture repository — tags on HEAD, nearest tag after a commit, annotated vs lightweight, no tags / no commits, detached HEAD, `overriddenBranchName`, `sinceCommit` exclusivity, pattern filtering, `taggedCommits` ordering, ahead/behind, legacy tag naming, monorepo paths. `ScmBackendIntegrationTest` covers the DSL flag, the Gradle property and an unknown backend value.

### Open questions

Fully understood if this does not fit given the v2 direction in #751 — no hard feelings if you would rather not grow v1.

Two things I am happy to change:

1. **The `repository.backend` DSL property.** As offered in #1075, I can drop it entirely and gate this on `-Prelease.scmBackend` only, so the public API does not grow at all. I left it in for discoverability, but say the word and I will strip it.
2. **Caching `GitRepository` results instead**, per the discussion in #182 — that would help both backends and cut more than this does (resolving one version currently issues ~70 git operations). I did not go there because it changes staleness semantics after `release` creates a tag.

### DIFF
--- a/docs/configuration/repository.md
+++ b/docs/configuration/repository.md
@@ -44,4 +44,27 @@ you do not want to set this in your build script:
 The command line flag **will** override the build script even if the
 script has explicitly set the `pushTagsOnly` flag to false.
 
+## Repository backend
+
+By default `axion-release` reads the repository with **JGit**. On very
+large repositories the tag walk and the working tree status scan
+performed during configuration can take seconds. In that case you can
+switch reads to the native `git` executable:
+
+    scmVersion {
+        repository {
+            backend.set("nativeGit")
+        }
+    }
+
+A command line flag, `release.scmBackend`, is also available:
+
+    ./gradlew currentVersion -Prelease.scmBackend=nativeGit
+
+Only reads use the `git` executable. Tagging, committing, fetching and
+pushing still go through JGit, so authorization keeps working exactly
+the same way - see [authorization](authorization.md).
+
+The native backend requires `git` to be available on `PATH`.
+
 See [authorization](authorization.md) for authorization options.
--- a/src/integration/groovy/pl/allegro/tech/build/axion/release/ScmBackendIntegrationTest.groovy
+++ b/src/integration/groovy/pl/allegro/tech/build/axion/release/ScmBackendIntegrationTest.groovy
@@ -1,0 +1,51 @@
+package pl.allegro.tech.build.axion.release
+
+import org.gradle.testkit.runner.TaskOutcome
+
+class ScmBackendIntegrationTest extends BaseIntegrationTest {
+
+    def "should resolve version with native git backend configured in build file"() {
+        given:
+        buildFile('''
+            scmVersion {
+                repository {
+                    backend = 'nativeGit'
+                }
+            }
+        ''')
+
+        runGradle('release', '-Prelease.version=1.0.0', '-Prelease.localOnly', '-Prelease.disableChecks')
+
+        when:
+        def result = runGradle('currentVersion')
+
+        then:
+        result.output.contains('1.0.0')
+        result.task(":currentVersion").outcome == TaskOutcome.SUCCESS
+    }
+
+    def "should resolve version with native git backend configured by gradle property"() {
+        given:
+        buildFile('')
+
+        runGradle('release', '-Prelease.version=1.0.0', '-Prelease.localOnly', '-Prelease.disableChecks')
+
+        when:
+        def result = runGradle('currentVersion', '-Prelease.scmBackend=nativeGit')
+
+        then:
+        result.output.contains('1.0.0')
+        result.task(":currentVersion").outcome == TaskOutcome.SUCCESS
+    }
+
+    def "should fail on unknown backend"() {
+        given:
+        buildFile('')
+
+        when:
+        def result = runGradleAndFail('currentVersion', '-Prelease.scmBackend=mercurial')
+
+        then:
+        result.output.contains("Unsupported scm backend 'mercurial'")
+    }
+}
--- a/src/main/groovy/pl/allegro/tech/build/axion/release/domain/RepositoryConfig.groovy
+++ b/src/main/groovy/pl/allegro/tech/build/axion/release/domain/RepositoryConfig.groovy
@@ -10,6 +10,7 @@ import org.gradle.api.tasks.InputFile
 import org.gradle.api.tasks.Optional
 import org.gradle.api.tasks.PathSensitive
 import org.gradle.api.tasks.PathSensitivity
+import pl.allegro.tech.build.axion.release.domain.scm.ScmBackend
 
 import javax.inject.Inject
 
@@ -26,6 +27,7 @@ abstract class RepositoryConfig extends BaseExtension {
     private static final String OVERRIDDEN_IS_CLEAN = 'release.overriddenIsClean'
     private static final String DISABLE_SSH_AGENT = 'release.disableSshAgent'
     private static final String FETCH_TAGS_PROPERTY = 'release.fetchTags'
+    private static final String SCM_BACKEND_PROPERTY = 'release.scmBackend'
 
 
     @Inject
@@ -33,6 +35,7 @@ abstract class RepositoryConfig extends BaseExtension {
         pushTagsOnly.convention(false)
         type.convention("git")
         remote.convention("origin")
+        backend.convention(ScmBackend.JGIT.id)
 
         customUsername.convention(gradleProperty(USERNAME_PROPERTY))
         customPassword.convention(gradleProperty(PASSWORD_PROPERTY))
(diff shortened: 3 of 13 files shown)


FireHook:
Hey @bgalek, when you get a chance — no rush.

It is opt-in, default stays JGit, so nothing changes unless you turn it on. Writes still go through the existing JGit path (tagging, push, fetch, auth all untouched), only reads are native. CI is green across the whole matrix.

On our repo (169k commits, 745 tags) it takes configuration from 24.5s on current `main` down to 19.0s.

And as I said in #1075 — genuinely fine if this does not fit given the v2 plans. Happy to drop the DSL property and leave only `-Prelease.scmBackend`, or to close it if you would rather not grow v1 at all.

